PR #111, opened 2015-01-09
Improve contacts controller

purezen:
Hi guys,
I am keen to contribute to the project. And have tried starting by improving some code in the contacts controller.
Please review my work.
Thx :)

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -1,5 +1,5 @@
 class ContactsController < ApplicationController
-  before_filter :authenticate_user!
+  before_action :authenticate_user!
 
   def new
     @message = Message.new
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -4,11 +4,8 @@ class ContactsController < ApplicationController
   def new
     @message = Message.new
     @subject = default_subject
-    user = User.find_by_id(params[:user_id])
-    if user.nil?
-      redirect_to root_path and return
-    end
-    @user = user
+    @user = User.find params[:user_id]
+    redirect_to root_path if user.nil?
   end
 
   def create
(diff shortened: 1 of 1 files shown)


hgducharme:
Looks good to me, however @eriktrautman is going to have to review and merge it in if he chooses to do so.

eriktrautman:
Thanks.  These changes don't really make sense to me -- see inline comments for details

app/controllers/contacts_controller.rb
```diff
@@ -1,37 +1,33 @@
...
+    redirect_to root_path if @user.nil?
   end
 
   def create
-    @user = User.find_by_id(params[:user_id])
+    @user = User.find params[:user_id]
     @message = Message.new({
         :body => params[:body].html_safe,
-        :subject => default_subject,
+        :subject => @subject,
```
eriktrautman: Where is this being set?  Why use an instance var?

app/controllers/contacts_controller.rb
```diff
@@ -1,37 +1,33 @@
 class ContactsController < ApplicationController
-  before_filter :authenticate_user!
+  before_action :authenticate_user!
+  before_action :set_subject
```
eriktrautman: Why is this being run as a filter?

app/controllers/contacts_controller.rb
purezen: Ah, looks like I missed this! Have fixed this in another commit on this PR.
Have used an instance variable since the new method assigns the message text to @message,
    @message = default_subject.

app/controllers/contacts_controller.rb
purezen: I am commenting on this PR as a whole. Let me know if it makes sense.

purezen:
So, one thing that I am trying to do here is improve the functionality of assigning the @subject. It is currently being done both in new and create methods explicitly.
I think the same can be done using the callback. The method would set @subject for both the methods (which I have 'just' assigned in the set_subject method).
Also, the method can be made private because it is only being used in the same controller.

PR #114, opened 2015-01-21
Fix Facebook like button display issue

chrisdziewa:
Adds 25px of width to Facebook button on front page.


herminiotorres:
:shipit:

PR #118, opened 2015-02-26
add js studygroup

mattphoto:
that was weird. not sure y ds_store got included. the js studygroup is open js studygroup with no specific agenda, so I didn't see a need to put a calendar up. I plan to have it every week for the foreseeable future.


eriktrautman:
I'll send out a flash notice about the group when you post the link to the hangout

PR #152, opened 2016-02-29
Remove the folders `script` and `plugins`.

herminiotorres:
- These folders were a version of the old `rails`


muzfuz:
Thanks, good catch.

PR #431, opened 2017-04-21
convert fixed-widths

codyloyd:
I didn't fix every problem.. but I fixed the overflow on widths < 350px


arun1595:
Thanks @codyloyd

PR #484, opened 2017-05-01
rename conflicting background properties in styleguide.scss

105ron:


codyloyd:
yeah.. good deal.. MErging

PR #560, opened 2017-05-26
Update course descriptions and short descriptions.

Demo318:
Consolidate course descriptions from paragraphs into just a few lines.


db/seeds.rb
```diff
@@ -1836,8 +1836,8 @@ def require_updates?(real_attrs, seed_attrs)
...
   title_url: "Javascript and jQuery".parameterize,
   teaser: "Make Your Websites Really Dance",
-  brief_desc: "The last component in your web development arsenal is the ability to make your front ends dynamically react to the user.  In this course you'll learn how to use Javascript and jQuery to interact with the DOM and make things like dropdown menus, parallax scroll effects, single page applications, and even games.  It will tie together everything you've learned already into one awesome package.",
-  description: "Javascript is taking over the world right now -- web applications are becoming more and more front-weighted.  While it used to be mostly a tool for animating drop down menus and validating form inputs, Javascript (with help from jQuery's magic) has grown into a way to produce full featured front end applications including single-page web apps.",
+  brief_desc: "Make dynamic, interactive webpages with JavaScript and jQuery.",
+  description: "JavaScript has become an essential piece of the modern website. This course will cover advanced principles including AJAX, APIs, and JavaScript frameworks. Create an incredible user experiences for your websites.",
```
arun1595: Create incredible user experiences ...

Omit 'an'

KevinMulhern:
Hey @Demo318, thanks for putting this together 😄 

I've read through the updated descriptions and think they are really good 👍 

After reading through these while trying to put my self in the shoes of someone knew to development and Odin. I think we should take a slightly different approach to this though. I think theres 2 things we want to communicate with the descriptions.

1. The courses at the moment are designed to be taken in order, if we could communicate how each course fits in to the overall picture that would be awesome.

2. I think we should emphasise what the user will be able to do by the time they complete each course.

So with these in mind what about descriptions like these?

Web Development 101
"To start things off, in this course you will get a quick hands-on overview for all the different parts of full stack development you'll use to build fully functioning web apps."

Ruby Programming
"Now that you've got a got overview of how all the different pieces of web development fit together. Its time to deep dive into learning programming using Ruby, a versitile, powerful and beginner friendly language designed for programmer happiness."

...

This is definitely up for discussion though, what do you think?

Demo318:
@KevinMulhern Thanks for the feedback!

I noticed elsewhere on the redesigned site that there's a "choose your path" emphasis. For this reason, I made descriptions as stand-alone as possible instead of one linking to the next. If we do want to emphasize the flow, we might want to also update the other mention so users are on the same page. Would you want this theme worked into both the short & long descriptions?

Also, I noticed in each 'new course' section of seeds.rb there's a "what you'll learn" and "what you'll build." I thought it best not to double-up with listing those items in the description as well as wherever that other info will end up (keeping it DRY).

If the "what you'll learn" and "what you'll build" info won't be visible alongside the description, I can certainly add it in.

KevinMulhern:
Hey @Demo318, ah of course I see 😄 

We plan on offering different paths in the future, we were very optimistic about how soon we could get that done though. So Ada designed that paths text on the homepage. We've recently removed that text from the homepage https://github.com/TheOdinProject/theodinproject/issues/553 it hasn't been deployed to the staging server yet though.

The  "what you'll learn" and "what you'll build." stuff in the seeds file was used in the old design. We won't be using them at all in the new design. I forgot about those, I must create a new issue to remove them from the courses model. Thanks for bringing it to our attention 😄

Demo318:
@KevinMulhern I appreciate the context. Super helpful!

I'll refactor this over the weekend.

Short Descriptions: I'll work in the logical flow, to emphasize moving from one subject to the next.

Long Descriptions: I'll add in the what you'll learn and what you'll build.

Thanks!

KevinMulhern:
Hey @Demo318 I'm not sure if its worth spending time on the long descriptions, those are only used on individual [course pages](http://localhost:3000/courses/web-development-101) we might be getting rid of them soon. We need to discuss that in the next design review though.

I think focusing on the brief descriptions is best for this ticket. 

Some sort of combination of the flow and what the student will be capable of when they complete each course is how I envisioned these descriptions.

This is the way I approached it
Imagining you're a new user to the site, as a new user you'll probably be exploring to see if the curriculum is for you are not. From that point of view you'll want to know how each part of the curriculum fits together and even more importantly what you will expect to be able to do by the end of the curriculum.

Maybe this is the wrong way to think about it, I personally think the descriptions have more benefit to new users than existing users. As existing users will already have a good idea of whats in the curriculum.

Demo318:
@KevinMulhern Got it.

With focusing on just the brief descriptions, can you give me a ballpark character count? How short is too short? How long is too long? This will help me make a block of text that fits well with the rest of the layout.

KevinMulhern:
I think limiting it to about 3 lines would be ideal 😄 that will lower the resistance of users to read them and look better on mobile.

Demo318:
@KevinMulhern Cool. ~225 characters looks like this. 

capture


Is that what you envision?

Demo318:
Thanks all for the feedback. Updated descriptions based on feedback from @KevinMulhern. I think JavaScript is still not exactly where it needs to be, but I hope we're close. Here are what they look like in the current design:

image
image
image
image
image
image

KevinMulhern:
These are absolutely perfect @Demo318 😍 

Thank you for doing this and all your patience with me 😄 

LGTM 👍 👍 👍 👍

PR #658, opened 2017-07-31
Allow students to add, edit and destroy projects

arun1595:
Tasks done:
- Allow students to add, update and destroy project submissions
- Allow students to vote on other students' projects
- Allow students to report other students' spammy submissions.

**Note**: Before deploying this feature, we've to move all the existing project submissions from the lesson README files into the `projects` table. Also, we need to setup ENV variables to post messages to our gitter chatroom.


app/controllers/lessons_controller.rb
```diff
@@ -9,9 +9,8 @@ def show
   private
 
   def set_user
-    if user_signed_in?
-      @user = User.includes(:completed_lessons).find(current_user.id)
-    end
+    return unless user_signed_in?
```
KevinMulhern: can we change this back to the way it was please? It reads better with an if imo.

app/controllers/projects_controller.rb
```diff
@@ -1,66 +1,72 @@
...
+    @projects = latest_projects
   end
 
   def update
-    @project.update(project_params)
+    @project_updated = @project.update(project_params)
+    @decorated_project = ProjectDecorator.new(@project)
+  end
+
+  def report
```
KevinMulhern: To keep things restful I think we should create a reports controller with a create action for this.

app/controllers/projects_controller.rb
```diff
@@ -1,66 +1,72 @@
...
+  def all_projects
+    Project.all_submissions(@lesson.id).order(updated_at: :desc)
   end
 
-  def set_recent_submissions
-    @submissions = Project
-                   .all_submissions(@lesson.id)
-                   .where.not(user_id: current_user.id)
-                   .limit(10)
+  def new_project(parameters = {})
```
KevinMulhern: Can we change this to use tap please?
```
current_user.projects.create!(parameters).tap do |project|
    project.lesson_id = @lesson.id
end
```

app/controllers/projects_controller.rb
```diff
@@ -1,66 +1,72 @@
...
     @course = CourseDecorator.new(@lesson.course)
+    @user = current_user
+    @project = @projects.where(user_id: current_user.id).first if current_user
   end
 
   def create
     @project = new_project(project_params)
-    @project.save
-    set_recent_submissions
+    @project_saved = @project.save
```
KevinMulhern: If this is only for a conditional check in the create.js view. We can get rid of this and do `@project.pesisted?` in the js view instead.

app/controllers/projects_controller.rb
```diff
@@ -1,66 +1,72 @@
...
+    @projects = latest_projects
   end
 
   def update
-    @project.update(project_params)
+    @project_updated = @project.update(project_params)
```
KevinMulhern: Same for this one, It's not the controllers responsibility to do things like this.

I would add a new method to the ProjectDecorator

```
def updated?
   project.updated_at != project.created_at
end
```

app/services/gitter.rb
```diff
@@ -0,0 +1,33 @@
+require 'json'
+
+module Gitter
+  class Client
+    include HTTParty
```
KevinMulhern: Can we use Faraday instead please? 

Faraday already is a dependency of a few of our other gems so it makes sense to use it over adding another similar gem.

app/services/project_reporter.rb
```diff
@@ -0,0 +1,40 @@
...
+    <<~MESSAGE
+      @/all Hi Odin mods :wave:, received a report from #{reporter.username} (#{reporter.email}) on the following project submission:
+      **Project ID**: #{project.id}
+      **Project Owner**: #{project.user.username} (#{project.user.email})
+      #{report_message}
+    MESSAGE
+  end
+
+  def report_message
+    return unless description.present?
```
KevinMulhern: Can we use an if here instead please?

app/assets/stylesheets/components/projects.scss
```diff
@@ -1,212 +1,91 @@
...
-  padding: 10px 0;
-  transition: all 0.2s ease-out;
-
-  &:hover {
-    background: $white-smoke;
-  }
-
-  .button_to {
-    display: inline;
+.project-form-container {
```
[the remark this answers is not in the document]
arun1595: Sure, let me do it.

app/services/project_reporter.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def self.report(project:, reporter:, description:)
+    new(project, reporter, description).report
+  end
+
+  def report
+    gitter_client.send_message(text: message, room_id: ENV['MODERATORS_CHATROOM_ID'])
+    true
+  rescue => e
```
KevinMulhern: Can we rescue the specific error thats likely to occur here please? rescuing everything is quite dangerous.

app/services/project_reporter.rb
```diff
@@ -0,0 +1,40 @@
...
+    false
+  end
+
+  private
+
+  attr_reader :project, :reporter, :description
```
KevinMulhern: The style we use in every other plain old ruby object is to have the attrs at the top of the class. Can we do the same here for consistency please?

KevinMulhern:
A couple of design problems I've noticed by using the feature:

* The toast when you upvote is unnecessary imo, the animation on the heart is enough.
* The modals should have an "X" to close them. Clicking outside of them to close them isn't intuitive in the report modal.
* The report link should be more obvious, we should talk to @adachiu about this.
* We should use a different color for the cancel button in the add submission modal
* The add submission button is the wrong place at the bottom of the projects container. It should be at the top.

screen shot 2018-01-15 at 21 48 40

app/services/project_reporter.rb
```diff
@@ -30,11 +30,11 @@ def message
   end
 
   def report_message
-    return unless description.present?
+    return if description.blank?
```
KevinMulhern: Sorry I should have been more clear with this one

```
if description.present
     "**Reporter's message**: #{description}"
end
```

Returning early from a method isn't as readable as a condition.

app/views/projects/_form.html.erb
```diff
@@ -1,25 +1,38 @@
 <div class="project-form-container">
-  <h3 class="text-primary text-center project-form-title">Upload your project</h4>
-  <div id="message"></div>
+  <h1 class="text-center project-form-title accent">Upload your project</h1>
 
   <%= form_for([lesson, project], remote: true) do |f| %>
     <div class="form-group">
```
KevinMulhern: Can we use our form component classes to style this form please?

app/views/projects/_project.html.erb
```diff
@@ -1,35 +1,42 @@
...
+  <div class="col-7 col-sm-9 col-md-3 col-lg-3 project__owner-name bold">
+    <% if project.user == current_user %>
+      <%= link_to '#',
+        data: {toggle: 'modal',
+          target: '#project-submission-modal'
+        } do %>
+       <%= project.owner_name.capitalize %>
+      <% end %>
+    <% else %>
+      <% if user_signed_in? %>
```
KevinMulhern: Moving this to its own partial would be better than having a nested conditional

app/views/projects/_report_form.html.erb
```diff
@@ -0,0 +1,26 @@
+<div class="report-project-form">
+  <div class="col-12">
+    <h1 class="text-center">Report project</h1>
+
+    <%= form_tag report_project_path, remote: true do |form| %>
```
KevinMulhern: Can we use our form component for the styles on this form please?

config/application.rb
```diff
@@ -8,10 +8,12 @@
 
 module Theodinproject
   class Application < Rails::Application
-    require Rails.root.join("lib/custom_public_exceptions")
+    require Rails.root.join('lib/custom_public_exceptions')
     config.exceptions_app = CustomPublicExceptions.new(Rails.public_path)
     config.middleware.insert_after ActionDispatch::Static, Rack::Deflater
 
+    config.action_controller.raise_on_unfiltered_parameters = true
```
KevinMulhern: Have you tested the login, signup and settings form still works after making this change?

spec/controllers/votes_controller_spec.rb
```diff
@@ -17,53 +12,54 @@
...
-        post :create, params: params
+        post :create, params: params, format: :js
         expect(project).to_not have_received(:upvote_for).with(user)
       end
     end
+
+    context 'returns bad request when user tries to vote his own project' do
+      before do
+        allow(controller).to receive(:current_user).and_return(:other_user)
+        post :create, params: params, format: :js
```
KevinMulhern: Can we remove this from the before do and put it into an `it` block please?

spec/controllers/projects_controller_spec.rb
```diff
@@ -1,386 +1,350 @@
...
+          it 'renders the destroy template' do
+            allow(student_project).to receive(:destroy)
+            expect(subject).to render_template(:destroy)
+          end
+        end
+      end
+    end
+  end
+
+  def stub_current_user_with(user)
```
KevinMulhern: These methods for stubbing aren't needed, they make the tests borderline unreadable imo. 

You can define a before do in the top level describe for this test and then override the stubs we need to change for each method in before do blocks within the method describe scopes.

This is the way we do it in every other test in the app.

spec/controllers/projects_controller_spec.rb
```diff
@@ -1,386 +1,350 @@
...
+          allow(student_project).to receive(:save).and_return(true)
+          allow(controller).to receive(:latest_projects).and_return(projects)
+          expect(subject).to render_template(:create)
+        end
+      end
+
+      context 'invalid values' do
+        subject { post :create, params: invalid_params, xhr: true }
+
+        it 'renders the create template' do
```
KevinMulhern: This  functions pretty much the same as when valid params are passed in so we can remove this test. Invalid params should be tested in end to end testing imo.

config/application.rb
arun1595: Not sure how this change got checked in. Will remove it.

app/views/lessons/show.html.erb
```diff
@@ -26,18 +40,30 @@
...
+        </div>
       <% end %>
+
+      <div id="new-submission">
+        <%= render 'add_submission_button' if @project&.new_record? %>
+      </div>
     </div>
 
-    <%= render 'shared/bottom_cta',
+      <%= render 'shared/bottom_cta',
```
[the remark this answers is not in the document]
arun1595: Good catch, fixed,

spec/controllers/votes_controller_spec.rb
arun1595: :see_no_evil: 

Let me fix it.

app/assets/stylesheets/components/projects/project.scss
```diff
@@ -0,0 +1,116 @@
+.project {
+  align-items: center;
+  border-bottom: solid 1px $very-pale-grey;
+  padding: 10px 0;
+  transition: all 0.2s ease-out;
+
+  &:hover {
+    background: $white-smoke;
+
+    & .project__report-link {
```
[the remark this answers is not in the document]
arun1595: @KevinMulhern Currently, the report link is only displayed when the project component is hovered.

Should we instead always display the report link in the project component?

Let me know what you think.

app/assets/stylesheets/components/projects/project.scss
```diff
@@ -0,0 +1,116 @@
...
+      display: inline;
+    }
+  }
+
+  .button_to {
+    display: inline;
+  }
+
+  &__votes {
```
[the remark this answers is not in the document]
arun1595: Done.

app/assets/stylesheets/components/projects/project.scss
```diff
@@ -0,0 +1,116 @@
...
+    .heart {
+      padding: 1em;
+    }
+
+    .fa { 
+      font-size: 2rem;
+    }
+  }
+
+  &__vote-button {
```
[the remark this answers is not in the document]
arun1595: Done.

app/assets/stylesheets/components/projects/project.scss
```diff
@@ -0,0 +1,116 @@
...
+
+    @include media-breakpoint-down(sm) {
+      justify-content: center;
+    }
+
+    @include media-breakpoint-only(md) {
+      padding: 0;
+    }
+
+    @include media-breakpoint-down(sm) {
```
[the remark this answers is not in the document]
arun1595: Done.

app/assets/stylesheets/components/projects/project.scss
```diff
@@ -0,0 +1,116 @@
...
+    .live-preview-button {
+      margin-left: 10px;
+    }
+  }
+
+  &__button {
+    background: $whitish-grey;
+    border-radius: 3px;
+    box-shadow: none;
+    color: $black;
```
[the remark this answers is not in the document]
arun1595: Ah, I changed it to black because of the low contrast ratio. I'll turn it to white for now. 

Lets address this in a separate PR.

app/assets/stylesheets/components/projects/submissions.scss
```diff
@@ -7,24 +7,51 @@
...
 
+  &__instructions {
+    @include media-breakpoint-down(sm) {
+      margin-top: 1rem;
+    }
+  }
+}
+
+#index-submissions {
   .pagination {
```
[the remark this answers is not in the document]
arun1595: Done.

app/controllers/lessons_controller.rb
arun1595: Done.

app/controllers/projects_controller.rb
arun1595: Done.

app/controllers/projects_controller.rb
arun1595: I am not sure if the logic to find a record is updated is right. Suppose the first update worked, in this case, `updated_at` field would get updated and won't be equal to the `created_at` field. so, the `updated?` method will return `true` as expected.

Lets say, the second attempt to update the project failed. In this case, `updated?` should return `false but ``project.updated_at != project.created_at` will return true which is a false positive.

app/controllers/projects_controller.rb
arun1595: I was initially using `tap` but found it harder to mock it in the tests. So, I resorted to the traditional way of doing. 

I'll try again to see if I can figure out how to mock it properly in the controller tests.

app/controllers/projects_controller.rb
KevinMulhern: That would suggest a flaw in the way we are doing things. Perhaps we should rethink things in the controller action for this.

arun1595:
@KevinMulhern I have addressed your review comments, please take another look when you get a chance.

Thanks!

KevinMulhern:
Hey @arun1595, can you deploy this to staging please?

app/assets/stylesheets/components/projects/modals.scss
```diff
@@ -0,0 +1,7 @@
+#project-submission-modal,
+#delete-project-modal,
+#project-report-modal {
+  .modal-dialog {
+    max-width: 920px;
```
KevinMulhern: Could this be added to the `modal_main` component instead?

app/assets/stylesheets/components/projects/form.scss
```diff
@@ -0,0 +1,21 @@
...
+
+  &__button {
+    width: 100%;
+
+    @media (min-width: 500px) {
+      width: 325px;
+    }
+  }
+
+  &__submit-btn {
```
KevinMulhern: I've been playing with the styles of this button on staging 😄 

The styles on this class don't seem to have any effect. Is this needed?

app/assets/stylesheets/components/projects/new_submission.scss
```diff
@@ -0,0 +1,16 @@
+.new-submission {
```
KevinMulhern: I'm not sure if this belongs in its own component. It's very specific and seems like it will always be an element of the submission component. Whats your thoughts?

app/assets/stylesheets/components/lesson/new_submission_button.scss
```diff
@@ -0,0 +1,5 @@
+@include media-breakpoint-up(lg) {
```
KevinMulhern: I don't think this belongs in its own component, imo we should wrap the button in an element and let that take care containing the size of the button. Checkout the contributing page buttons to get an idea of how we do that elsewhere https://www.theodinproject.com/contributing

app/views/lessons/show.html.erb
```diff
@@ -26,14 +26,26 @@
...
+        <%= @lesson.title %>
+      </h3>
+
       <div class="lesson-button-group">
         <%= render 'lesson_buttons', lesson: @lesson, course: @lesson.course, user: @user %>
       </div>
 
       <% if @projects && ProjectSubmissionFeature.enabled? %>
-        <%= render 'submissions' %>
+        <div id="submissions">
```
KevinMulhern: Do we need this div with the submissions id?  it doesn't seem to serve any purpose other than being used a selector for the js. If a selector is all thats needed here can we use the `submissions-container` instead please?

KevinMulhern:
Hey @arun1595, I'm going to deploy this :shipit:

PR #804, opened 2017-10-28
Remove wd101/testing_basics from seed

leosoaivan:
TheOdinProject/ruby_course#4

I'm not sure why the deletion looks like it's happening mid-lesson.


arun1595:
> I'm not sure why the deletion looks like it's happening mid-lesson.

Don't worry about that, it is how the diff algorithm `git` uses works :smile:

PR #837, opened 2017-12-11
DESTROY current JS content and MERGE new stuff

codyloyd:
well friends... here it is, for better or worse.

I have a handful of extra lessons that I intend to add in the coming days... but this feels like a viable minimal starting point...

there are a few 'commented out' sections in the seeds file. those are lessons that are upcoming.. I've left them in there for now to reduce the headache of adding them in as they are finished.

Be aware that this PR tosses a majority of the current JS curriculum and replaces it with a bunch of all new content.


:Ex
```diff
@@ -0,0 +1,95 @@
+html {
```
KevinMulhern: whats this file for? 😄

KevinMulhern:
👍 LGTM
We might need to remove the lessons that have been removed through the rails console. The seeds file uses the `title` attribute to find an existing lesson and update it. So if a title is changed it will create a new lesson and leave the old one.

:Ex
codyloyd: for nothing i'm retarded pls ignore

codyloyd:
I added a few more lessons.. this thing is almost done
localhost_3000_courses_javascript_lessons_es


@KevinMulhern yeah, we'll have to do that unfortunately.  I have been resetting the db everytime I update anything.. which is probably not quite an option for the production DB lol

codyloyd:
I'm a little sad to see snake and minesweeper go......... but I'm not really sure where they'd fit into the curriculum.

KevinMulhern:
No worries about the left over lessons, It should tell us which ones aren't being used anymore after the seed command runs.

I'll merge this tomorrow evening if thats ok? I'm won't be back home until late this evening 😭

codyloyd:
that works for me.. I'm not in a big hurry

PR #841, opened 2017-12-26
add link the lesson badge and course title on lesson pages

105ron:
Add a link back to the course overview page on the course badge and course heading on the lesson pages.

### DIFF
--- a/app/views/lessons/show.html.erb
+++ b/app/views/lessons/show.html.erb
@@ -2,10 +2,12 @@
 
 <div class="container">
     <div class="col-lg-8 offset-lg-2">
-      <div class='lesson-header'>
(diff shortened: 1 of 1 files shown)


app/views/lessons/show.html.erb
```diff
@@ -2,10 +2,12 @@
 
 <div class="container">
     <div class="col-lg-8 offset-lg-2">
-      <div class='lesson-header'>
+    <div class='lesson-header'>
```
arun1595: Nit. Can we use double quotes for html attribute values?

arun1595: Nit. Indentation is off.

PR #922, opened 2018-07-17
Spelling correction for seeds

105ron:
Environment was incorrectly spelt in the seeds.


arun1595:
@105ron Thanks for the fix!

PR #925, opened 2018-07-18
update seeds for js frameworks project to style it as a project

105ron:
The new JS frameworks project isn't listed as a project in the seeds. This is my fix.


arun1595:
@105ron Good catch :+1: 

LGTM :shipit:

PR #953, opened 2018-11-25
Remove gitter

codyloyd:
this pr removes mentions of gitter all around the site and replaces them with discord.. I'm fairly certain that I got all the links and buttons

there are a couple of analytics things and related css classes that I left, mostly because it doesn't matter and I didn't want to break anything.

I pushed all the changes to https://staging-odin-project.herokuapp.com/ if you wanna SEE it.

### DIFF
--- a/app/views/shared/_sidecar.html.erb
+++ b/app/views/shared/_sidecar.html.erb
@@ -1,6 +1,6 @@
 <div class="floating-dropdown">
-  <div class="testing floating-dropdown__item floating-dropdown--hidden-item text-center" data-toggle='tooltip' data-placement='left' title='Chat'>
-    <%= link_to chat_link, target: :_blank, rel: 'noreferrer', class: "testing" do %>
+  <div class="floating-dropdown__item floating-dropdown--hidden-item text-center" data-toggle='tooltip' data-placement='left' title='Chat'>
+    <%= link_to chat_link, target: :_blank, rel: 'noreferrer' do %>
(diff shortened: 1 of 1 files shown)


app/helpers/application_helper.rb
```diff
@@ -122,7 +122,7 @@ def faq_items
       {
         question: 'How can I get in touch?',
         answer:
-        "<p>If you have questions about The Odin Project itself, feel free to <a href='mailto:contact@theodinproject.com'>contact us</a>.  There is also a friendly Odin community in our <a href='https://gitter.im/TheOdinProject/theodinproject'>Gitter chat rooms</a></p>"
+        "<p>If you have questions about The Odin Project itself, feel free to <a href='mailto:contact@theodinproject.com'>contact us</a>.  There is also a friendly Odin community in our <a href='https://discord.gg/hvqVr6d'>Discord chat rooms</a></p>"
```
arun1595: What do you think of using the `chat_link` method instead of hardcoding the URL?

codyloyd: no reason not to.... good catch

codyloyd: well maybe there's a reason not to

codyloyd: looking at that file.. those are just strings that get turned into html at some point.. I don't think I can do a `link_to` in there... and this doesn't really hurt anything.

Issue #960, opened 2019-01-07
Navbar out of alignment


mindovermiles262:
While browsing the site with my screen split, the navbar is out of alignment:

![screenshot_2019-01-07 dashboard the odin project](https://user-images.githubusercontent.com/14863257/50800411-f10afe00-1294-11e9-839b-b3b39d94e4ee.png)

Also effects users that are not logged in.

Browser Resolution: 952x932
Effects: Chrome, FireFox (Not tested on others)
OS: Linux

dos077:
I found the issue in css. The ::after element in .dropdown-toggle class uses vertical-align: 0.255em to align itself. That throws off the padding calculation of <a.nav-link> in tablet mode. Testing in chrome using position:relative and top: -0.255em seems to fix the issue by isolation the vertical offset to only the ::after element. I cannot find the css file in the odinproject repo though.

dos077:
Pull request sent. I am in the middle of the "Using Git in the Real World" section. I tried to follow all the steps, hopefully there's no screw up.

PR #966, opened 2019-01-19
Fix/fixed navbar out of alignment

mt9304:
Fixed the navbar out of alignment issue  #960. Reducing the padding of the navbar area to 5vh from 7.5vh seems to resolve it. Mobile menu still looks okay. Would this be breaking any other menu items?

tablet
mobile


arku:
@mt9304 Thanks for your PR. I have merged https://github.com/TheOdinProject/theodinproject/pull/970 which removed the unnecessary padding and fixed the line-height (also removed the duplicate `line-height` declaration).

PR #974, opened 2019-02-02
Feature/tracks

mindovermiles262:


app/models/track.rb
```diff
@@ -1,6 +1,6 @@
 class Track < ApplicationRecord
   has_many :users
-  has_many :track_courses
+  has_many :track_courses, dependent: :destroy
```
arku: What do you think about using `dependent: :delete_all` instead?

I don't think we have any callbacks on `TrackCourse`. So there's no need to instantiate objects and delete them one by one. We can delete them all with a single `DELETE` SQL query.

app/controllers/tracks_controller.rb
```diff
@@ -11,6 +11,6 @@ def index
   private
 
   def ordered_decorated_courses(track_courses)
```
arku: Why do we need to pass the `track_courses` as an argument?

Is there any reason not to use `@track.courses` inside `ordered_decorated_courses` method?

app/models/track.rb
arku: What do you think about using `dependent: :delete_all` instead?

I don't think we have any callbacks on `TrackCourse`. So there's no need to instantiate objects and delete them one by one. We can delete them all with a single `DELETE` SQL query.

db/seeds/08_full_stack_track.rb
```diff
@@ -7,8 +7,18 @@
   Track.delete_all
 end
 
-Track.create!(
+create_or_update_track(
   title: "Full Stack",
   description: "The OG Odin",
-  position: 1
+  position: 1,
+  courses: [
```
arku: @mindovermiles262 The track courses created don't have their `track_position`s set. Can you add those?

app/models/track.rb
mindovermiles262: That will work much better. We do not need to instantiate each object before deleting. Changed to use `delete_all`

PR #978, opened 2019-02-15
Feature/anchor links

morrowdavidson:
internallink
In the process of getting this feature up. I have the links in still need to add CSS to show link icon when you hover.


mindovermiles262:
@KevinMulhern @CouchofTomato

morrowdavidson:
I also refactored the lesson.js a bit so that it calls `getElements('.lesson-content h3')` just once.

morrowdavidson:
Actual screenshot of the `:hover` effect:
internallink02

app/assets/stylesheets/global.scss
```diff
@@ -115,3 +115,21 @@ textarea {
   font-family: $body-fonts;
   padding: 0.7rem 1rem 0.7rem 1rem;
 }
+
+a.internal-link {
+  color: $text-primary;
+  text-decoration: none;
+  padding-left: 1.1em;
+  margin-left: -1.1em;
```
KevinMulhern: why do we need a minus margin here?

app/assets/stylesheets/global.scss
```diff
@@ -115,3 +115,21 @@ textarea {
   font-family: $body-fonts;
   padding: 0.7rem 1rem 0.7rem 1rem;
 }
+
+a.internal-link {
```
KevinMulhern: Since internal link only works within lesson files can we move this class into the lesson content component style sheet please? 
https://github.com/TheOdinProject/theodinproject/blob/master/app/assets/stylesheets/components/lesson/lesson_content.scss

app/assets/stylesheets/global.scss
```diff
@@ -115,3 +115,21 @@ textarea {
...
...99,3.916 c-0.339,0.521-0.539,0.982-0.87,1.313c-9.028,9.003-17.849,18.229-27.166,26.924c-11.683,10.903-28.762,13.169-42.121,6.186 C-1.531,137.299-6.49,110.041,9.457,92.094c10.938-12.31,22.768-23.92,34.971-34.992c13.869-12.585,34.784-12.429,48.045-0.726 c4.964,4.381,5.809,10.165,2.121,14.516c-3.713,4.381-9.422,4.476-14.614,0.241c-6.634-5.41-15.895-6.054-22.348-0.023 C45.772,82.192,34.05,93.533,23.366,105.726c-8.563,9.773-3.426,24.162,8.543,26.604c6.834,1.395,12.53-1.291,17.329-5.984 C54.863,120.843,60.355,115.208,65.95,109.584z"/> <path fill="DarkGoldenRod" fill-rule="evenodd" clip-rule="evenodd" d="M86.186,42.461c-7.526-4.352-15.039-5.282-22.953-4.119 c0.112-0.363,0.112-0.658,0.26-0.805c9.298-9.255,18.249-18.895,27.993-27.653c17.782-15.984,44.714-12.009,55.879,7.778 c7.527,13.339,5.709,31.127-5.288,42.832c-10.943,11.648-22.257,22.996-34.001,33.833c-14.049,12.966-35.044,13.138-48.567,1.186 c-4.906-4.337-5.749-10.221-2.081-14.529c3.67-4.312,9.489-4.378,14.64-0.167c6.957,5.686,16.705,5.663,23.562-0.905 c10.913-10.455,21.71-21.071,31.943-32.183c9.317-10.115,5.511-24.887-6.728-27.945c-6.912-1.728-12.832,0.697-17.757,5.543 C97.397,30.929,91.864,36.692,86.186,42.461z"/> </g> </g> </svg>');
```
KevinMulhern: We already pull font awesome into the site, to avoid loading a third party image can we use that instead? https://fontawesome.com/icons/link?style=solid

KevinMulhern:
Hi @morrowdavidson, this looks really good. I left some minor feedback :)

app/assets/stylesheets/global.scss
morrowdavidson: that's a trick to get the background image to appear to the right

app/assets/stylesheets/global.scss
morrowdavidson: I was thinking the same but it works here too : 
internallink3

KevinMulhern: ah, very nice!

In that case I would probably put this in its own component style sheet.

app/assets/stylesheets/global.scss
morrowdavidson: I pulled in font awesome's .svg. As the css currently works because it changes the **background image** when you hover I can't use a font.

KevinMulhern: ah yes, I think a pseudo before selector would be needed. I won't hold this up with that though. We can always create a new issue for a small refactor.

KevinMulhern:
@arku can you take a look over this when you get a moment please?

arku:
@KevinMulhern Sure, let me take a look.

app/assets/javascripts/lessons.js
```diff
@@ -89,40 +99,53 @@ function isLessonPage() {
 }
 
 function constructLessonSections() {
-  var lessonHeadings = getElements('.lesson-content h3');
-  lessonHeadings.forEach(function(heading) {
-    const id = heading.getAttribute('id');
-    heading.removeAttribute('id');
+  lessonHeadings.forEach(buildScrollSpy);
+  uri = location.href.replace(location.hash,"");
```
arku: And please use single quotes to stay consistent.

arku: There are a few other places where single quotes should be used. I haven't commented in all those lines but please get them replaced.

morrowdavidson:
Thanks @KevinMulhern @arku for your input and quick comments! :)

KevinMulhern:
👍 Looks good to me

If your happy with it @arku, could you merge it please? :)

PR #1019, opened 2019-08-15
Upgrade Nokogiri

arku:
See https://github.com/sparklemotion/nokogiri/issues/1915


arku:
@CouchofTomato Thanks for the review!

Issue #1021, opened 2019-08-20
Profile card on Dashboard stretches


JoeVinten:
At different sizes the profile card stretches the image inside.

I believe this could be solved just by adding a `<div>` around in it. Happy to work on this.

![image](https://user-images.githubusercontent.com/33879422/63386773-bf662380-c39b-11e9-88d9-23775e278f4b.png)

PR #1022, opened 2019-08-20
added a div to the user profile card to ensure it stops stretching

JoeVinten:
I believe I've fixed this issue I was having (in chrome):

https://github.com/TheOdinProject/theodinproject/issues/1021


arku:
@CouchofTomato Thanks!

PR #1027, opened 2019-10-01
Resolve CVE-2015-9284

arku:
Follows the steps outlined in https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284#mitigating-in-rails-applications

Since we were already using `POST` methods for links that initiate the auth request phase, we didn't have much to do other than adding `omniauth-rails_csrf_protection` gem and ignoring the advisory CVE-2015-9284 in bundle audit task.


arku:
@CouchofTomato Thanks for the review and the merge 🙂

PR #1033, opened 2019-10-11
Add Rubocop Performance

CouchofTomato:
The rubocop post-install message now states performance cops will be
removed in Rubocop 0.68 and are now maintained in the Rubocop
Performance Gem. This adds the gem and requires the file in the rubocop
yml


arku:
@CouchofTomato You're on 🔥

Issue #1035, opened 2019-10-12
Duplicate `TrackCourse`s are created when seeding the database


arku:
When we run `rails db:seed` mutiple times, it looks like duplicate `TrackCourse`s are getting created.

Please see #1034 for more context.

CouchofTomato:
Thanks to @catalindrg this issue seems to be resolved.

mindovermiles262:
Do we need to go in and manually delete duplicates?

CouchofTomato:
@mindovermiles262

Yes. If you can handle it then great otherwise I’ll look when I’m home. It’s the courses linked to tracks 2 and 3 as far as I can tell

arku:
@mindovermiles262 The duplicates should be gone now.

PR #1036, opened 2019-10-12
Fix duplicate trackcourse creation in seed

catalindrg:
[Issue](https://github.com/TheOdinProject/theodinproject/issues/1035)


arku:
Thanks for the patch @catalindrg!

catalindrg:
Happy to help!

PR #1037, opened 2019-10-13
redirect /courses to /tracks

mindovermiles262:
Solves #1015 where /courses should point to /tracks

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -40,6 +40,7 @@ Rails.application.routes.draw do
   # Deprecated Route to Introduction to Web Development from external links
   get '/courses/introduction-to-web-development' => redirect('/courses/web-development-101')
 
+  get '/courses' => redirect('/tracks')
   resources :courses, only: %i(index show) do
     resources :lessons, only: :show
   end


arku:
@mindovermiles262 Isn't #1015 solved by https://github.com/TheOdinProject/theodinproject/pull/1028? 🤔

mindovermiles262:
I guess that depends on if we want to be able to access the legacy `/courses` uri

#1028 only changes the button link, whereas this PR will change it for any other buttons that may be hidden inside the application

arku:
@mindovermiles262 Right. But do we always want to redirect to `/tracks`?

mindovermiles262:
I'd say yes. 

`/courses` currently points to only Full Stack RoR but we want to show off the offerings we've worked so hard on.

arku:
> /courses currently points to only Full Stack RoR 

@mindovermiles262 I think we're displaying a list of all the courses there. But I think you're right. It makes sense to display the tracks instead of the all courses page.

arku:
@mindovermiles262 Let's see what others think about this change.

mindovermiles262:
ACK. Let's wait to hear from at least one other RE: if we should redirect `/courses` (displays Full Stack RoR course) to `/tracks` (Displays all available tracks)

arku:
> ACK

🤓

mindovermiles262:
Sounds good. Merging

PR #1045, opened 2019-11-19
fix #1021 issue, removed fixed height to cause profile card to not st…

### LINKED ISSUE #1021: Profile card on Dashboard stretches

bill-mansfield:
Fix #1021 issue
Removed fixed height to cause profile card to not stretch image (See screenshot below: green circle with "B" inside).

Profile card image will decrease in total size, but in my opinion this is a more aesthetic display of the image.

Screen Shot 2019-11-19 at 11 49 16 am

### DIFF
--- a/app/assets/stylesheets/components/dashboard/profile_card.scss
+++ b/app/assets/stylesheets/components/dashboard/profile_card.scss
@@ -11,7 +11,6 @@
 
   &__image {
     width: 100px;
-    height: 100px;
     background-color:#fff;
     border-radius:50%;
     margin-right: 2rem;


CouchofTomato:
@arku Have you been able to replicate this error?

arku:
@CouchofTomato Yeah.

CouchofTomato:
We do seem to have two PR's for this.

This one and https://github.com/TheOdinProject/theodinproject/pull/1022

If we merge this one we should close the other one

arku:
@CouchofTomato You're right. But this solution makes the profile picture smaller (~71px).

image

#1022 retains the height and width and still fixes the problem.

Which solution do you prefer?

CouchofTomato:
I would assume we want to retain the height and width?

arku:
@CouchofTomato Ah, the author had actually mentioned this.

> Profile card image will decrease in total size, but in my opinion this is a more aesthetic display of the image.

WDYT?

arku:
@CouchofTomato Got it. Let's close this one then.

CouchofTomato:
Yeah I'll close this because we haven't discussed with everyone whether the smaller image is better. We can merge the other and then consider changes to image size seperately.

PR #1050, opened 2019-12-20
Update Rack gem to 2.0.8

CouchofTomato:
Release notes can be found at
https://github.com/rack/rack/blob/master/CHANGELOG.md

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -265,7 +265,7 @@ GEM
     psych (3.1.0)
     public_suffix (3.0.3)
     puma (3.12.2)
-    rack (2.0.7)
+    rack (2.0.8)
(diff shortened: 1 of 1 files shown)


leila-alderman:
This seems like something that needs to also be updated in the `Gemfile`, not just `Gemfile.lock`, so that the change will remain after a fresh `bundle install`.

CouchofTomato:
It’s just a minor bump and rack isn’t listed in the Gemfile so it should be ok on a fresh bundle install and would install the latest version.

arku:
@leila-alderman Austin is right. When you do a `bundle install`, bundler will install the gems locked to the versions specified in `Gemfile.lock`.

We are good to go 🚀

PR #1055, opened 2020-01-05
Add line numbers to code blocks

### LINKED ISSUE #1054: Enable Line Numbers for Code Snippets
In the TOP lessons, code snippets are frequently referred to by line number. This is currently inconvenient, because line numbers are not shown. 

As far as I know we use kramdown for markdown conversion. And I found a potential solution to enable line numbers here: https://github.com/jekyll/jekyll/issues/4619 
```
kramdown:
    syntax_highlighter_opts:
        block:
            line_numbers: true
```

I have a setup of TOP locally, but I don't know where to insert that code.

arku:
### Description

Adds [Line numbers](https://prismjs.com/plugins/line-numbers/) plugin to Prism.

Fixes #1054 

### Screenshots

image

image

### DIFF
--- a/app/assets/javascripts/prism.js
+++ b/app/assets/javascripts/prism.js
@@ -1565,3 +1565,175 @@ Prism.languages.sql= {
 	'operator': /[-+*\/=%^~]|&&?|\|?\||!=?|<(?:=>?|<|>)?|>[>=]?|\b(?:AND|BETWEEN|IN|LIKE|NOT|OR|IS|DIV|REGEXP|RLIKE|SOUNDS LIKE|XOR)\b/i,
 	'punctuation': /[;[\]()`,.]/
 };
+
+/**
+ * Prism Line numbers plugin -  https://prismjs.com/plugins/line-numbers/
+ */
+(function () {
+
+	if (typeof self === 'undefined' || !self.Prism || !self.document) {
+		return;
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


nearmint:
This works great! Thank you @arku. 
One example, where code snippets are referred to by line numbers is here: https://www.theodinproject.com/courses/javascript/lessons/factory-functions-and-the-module-pattern

I think this is a good improvement.
Thanks again @arku

arku:
@nearmint Thanks for the review!

I just updated the PR description with a screenshot from `/courses/javascript/lessons/factory-functions-and-the-module-pattern`.

mindovermiles262:
Merging and pushing to staging

nearmint:
I notice two issues, first issue:
Selection_051
Taken from: https://staging-odin-project.herokuapp.com/courses/javascript/lessons/objects-and-object-constructors
As you can see, there is a line under line number 3 which is not numbered. That is probably because the code snippet doesn't have a high enough width to display the full line. I'm not sure if we need to fix these cases somehow (higher width, or a line break) or whether we want to leave this as-is.

2. In general I noticed that the line numbers are not in the center of the line, but rather start somewhat "under" the line. 
OnPaste 20200118-220648
I need your opinion whether that's acceptable.

dvik1950:
I think #1 isn't really an issue. One way to make it look better would probably be splitting long comments into multiple rows but it's kinda fine now as it is imo.

PR #1059, opened 2020-02-16
update username validation

scottrobel:
This will make the js validations the same for client and server-side
https://github.com/TheOdinProject/theodinproject/issues/1056


scottrobel:
It just makes sense for the front and back end validations to be the same.

PR #1065, opened 2020-03-11
changes FAQ to include extra information

zachmmeyer:
Small changes to FAQ.


I3uckwheat:
Thanks! This'll show when the site gets updated!

PR #1090, opened 2020-04-15
Add tooling to help make data informed decisions 

scottrobel:
this is a rake task I wrote to tell the average duration it takes to finish any given lesson. I hope this data will be used to improve the curriculum.
You can now see if a lesson is taking longer than it should.
by running 
`bundle exec rake user_data:print_completion_data`
you will see something like this: 
rake_results
If this works as intended I might try and make a page to visualize this data.


scottrobel:
If this works as intended I might try and make a page to visualize this data.

I3uckwheat:
@scottrobel Are you telling me that people only take 50 seconds to finish webdev101?

codyloyd:
@scottrobel can you provide a little more detail on what data you're grabbing and comparing here? I'm looking through the code and what I _think_ you're doing is interesting.. but it's been a long time since I wrote any Ruby so I'm not completely sure lol.

app/services/lesson_completion_data.rb
```diff
@@ -0,0 +1,72 @@
...
+        completion_durations[next_lesson] = ActiveSupport::Duration.build(average_duration_to_finish_next_lesson)
+        current_lesson_avg_completion_time = next_lesson_avg_completion_time
+        current_lesson = next_lesson
+      end
+    end
+    completion_durations
+  end
+
+  def lesson_completions_count
+    lesson_completions_count ||= LessonCompletion.group(:lesson_id).count
```
CouchofTomato, line 60: What is supposed to be happening here?

scottrobel:
> @scottrobel Are you telling me that people only take 50 seconds to finish webdev101?
Yes I don't have production data

scottrobel:
> can you provide a little more detail on what data you're grabbing and comparing here?

Sure thing. So This Calculates the average time between a user marking one lesson as complete and a user marking the next as complete. I can find this data from the "created_at" column of the lesson_completions table.
So this became the problem (How can I find the overall average distance between two sets of data(one lessons lesson_completions.created_at and the next lessons lesson_completions.created_at)). I found the most efficient way to find this data is to find the overall averages of both sets and find the distance between the two overall averages.
` LessonCompletion.all.group(:lesson_id).average('extract(epoch from created_at)')`
This code finds the average datetime every lesson was completed in epoch format.
To find the average time a lesson takes to complete
1. Need to find the average completion date the lesson before
2 You need to find the average completion date of the lesson you are calculating.
3.Need to find the difference between these two numbers
the method 
`known_completion_durations` takes the average completion date and finds the duration of each lesson (untill there are no lesson_completions for the next lessons(hence the name known))
Ask more questions if you don't full understand. I will be happy to answer.

app/services/lesson_completion_data.rb
scottrobel: > What is supposed to be happening here?
This is a method I added in to show the number of lesson_completions for each lesson.
it will output a hash like {43 => 213}(43 being the lesson_id)(231 being the amount of people who finished the lesson)
the ||= is there so it doesn't run the query twice.

scottrobel: > can you provide a little more detail on what data you're grabbing and comparing here?

Sure thing. So This Calculates the average time between a user marking one lesson as complete and a user marking the next as complete. I can find this data from the "created_at" column of the lesson_completions table.
So this became the problem (How can I find the overall average distance between two sets of data(lesson_one.lesson_completions.created_at and the next_lessons lesson_completions.created_at)). I found the most efficient way to find this data is to find the overall averages of both sets and find the distance between the two overall averages.
` LessonCompletion.all.group(:lesson_id).average('extract(epoch from created_at)')`
This code finds the average datetime every lesson was completed in epoch format.
To find the average time a lesson takes to complete
1. Need to find the average completion DateTime of the lesson before
2. You need to find the average completion DateTime of the lesson you are calculating.
3. Need to find the difference between these two numbers
the method 
`known_completion_durations` takes the average completion date and finds the duration of each lesson (untill there are no lesson_completions for the next lessons(hence the name known))
Ask more questions if you don't full understand. I will be happy to answer.

scottrobel: user_data
Here is a screenshot showing the output with the new data

CouchofTomato: > 
> 
> > What is supposed to be happening here?
> 
> This is a method I added in to show the number of lesson_completions for each lesson.
> it will output a hash like {43 => 213}(43 being the lesson_id)(231 being the amount of people who finished the lesson)
> the ||= is there so it doesn't run the query twice.

Why do you need the local variable?

scottrobel:
rake_results_weight
Added weight

.ruby-version
```diff
@@ -1 +1 @@
-2.6
+2.6.0
```
mindovermiles262, line 1: Please be sure to change this back

scottrobel:
Kevin requested I calculate the lesson durations like this. 

> Pass a date range into the rake task. for each course order the lessons and group each lesson pair, a lesson, and the one that comes afterward. Loop through the lesson pairs. Find all the users who have a lesson completion for the first lesson in the pair within the date range. If they haven’t completed the next lesson don’t do anything. Otherwise, if they completed the next lesson calculate the duration and store it against the lesson. finally average the durations

Most of the work is done in the database.
```
  def all_lesson_durations(lesson_id, last_lesson_id)
    durations =\
    @lesson_completions\
    .where(lesson_id: [lesson_id, last_lesson_id])\
    .group('lesson_completions.student_id')\
    .having("count(lesson_completions) = 2")\
    .pluck("max(extract(epoch from created_at)) - min(extract(epoch from created_at))")
  end
```

scottrobel:
@KevinMulhern Changes have been made.

scottrobel:
2 Major changes I just made on Kevin's request.
1. Omit all data but the average Lesson Duration
2. output everything to the logs instead of the console

scottrobel:
lesson_logs1
Here is what the logs look like after running the command.

PR #1250, opened 2020-07-23
Better cache and remove DB dependency on docker build

inductor:
- `MAINTAINER` is deprecated. Ref. https://docs.docker.com/engine/reference/builder/#maintainer-deprecated 
- `ADD` is anti-pattern and should be `COPY` unless there's any specific reason. Ref. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
- Having DB dependency in docker build is not "portable", which is what Docker should bring us

I also can add `docker-compose` for PostgreSQL not just for those who want to use Docker for this Rails app, but also who don't want to prepare DB on their local environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,21 +5,20 @@
 #
 
 FROM ruby:2.6-buster
-MAINTAINER Andy Duss <mindovermiles262@gmail.com>
+LABEL maintainer="Andy Duss <mindovermiles262@gmail.com>"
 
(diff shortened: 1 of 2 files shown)


CouchofTomato:
@mindovermiles262 Can you review this please.

mindovermiles262:
Sorry this took so long to review. Looks good! Thank you!

mindovermiles262 (approved):

PR #1295, opened 2020-08-25
Refactor and add context to project submission component

RobPando:
First, great job 🎉  I honestly think this was a great implementation and code. I only did minor refactoring and added the context to share some data that not all components needed.

To the context I added `userId`, `lesson` and `course` since these won't change unless user navigates to another lesson or course in which case a new component will be rendered anyways. The only data that was really being passed down multiple places without the need to, was `userId` but I included` lesson` and `course as well, it is redundant right now but it also doesn't cause any harm.

I tested this to the best of my knowledge, but it would be best to test this further as you have deeper knowledge of how exactly it is suppose to work and maybe I missed some crucial parts.

**Additionally:** besides the refactoring, I did catch 2 crucial things, one was that current user was able to flag their own project it seems silly but I hid that action, and the most important one is that all submissions were being shown so I added a scope to only show publicly enabled submissions.

Let me know what you think and if you have any questions whatsoever.

### DIFF
--- a/app/controllers/lessons_controller.rb
+++ b/app/controllers/lessons_controller.rb
@@ -18,7 +18,7 @@ class LessonsController < ApplicationController
   end
 
   def project_submissions
-   (current_users_submission + lesson.project_submissions.with_no_active_flags.limit(10)).uniq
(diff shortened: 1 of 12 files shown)


KevinMulhern (approved):
Looks great @RobPando 🌮 💪

PR #1306, opened 2020-08-30
Add active admin theme and flags admin panel

dvik1950:
#1125 :  Solves "Implement Active Admin Page Flag Page Which will Display Details about the Report."

I added two additional libraries for active admin, one of them is mostly UI stuff the other adds additional functionalities to the admin panel.

I changed the `action_taken` column to `taken_action` so the enum can be retrieved as `Flag.taken_actions` instead of `Flag.action_takens` which didn't seem right.

The Flag panel is pretty basic, let me know if I should change anything or add something else.


KevinMulhern (approved):
👍 Looks great to me @dvik1950 💪

PR #1407, opened 2020-10-14
Redirect tracks urls to paths

CouchofTomato:
Now that tracks have been renamed to paths this commit updates the routes file to redirect tracks to paths. This is to ensure anybody still trying to reach tracks doesn't find a missing page.


ZachBaird (approved):
Looks good to me, but Rubocop seems mad about the second line.

PR #1410, opened 2020-10-15
Adding a button to the paths show view to make switching paths more prominent.

ZachBaird:


ZachBaird:
I re-added the paths.scss just to throw some styling in there for the title margin and the new button.

I3uckwheat:
@ZachBaird Can you share screenshots of before and after?

ZachBaird:
@I3uckwheat Sure, I'll have it up later today.

ZachBaird:
@I3uckwheat Here is a screenshot:
image

Per the discussion earlier, there's just no button if you're on the enrolled path.

app/views/paths/show.html.erb
```diff
@@ -1,8 +1,14 @@
 <div class="container">
   <h1 class="text-center camel light curriculum-title"><%= @path.title %></h1>
+  <% if user_signed_in? %>
```
KevinMulhern, line 3: I think it might read a little better if these conditionals were combined into one:
`<% if  user_signed_in? && current_user.path != @path %>` 

whats your thoughts?

KevinMulhern (commented):
Just one little nitpick but otherwise looks great @ZachBaird, thanks for putting this together 💪

app/views/paths/show.html.erb
ZachBaird: I think it's a great idea! I'll have an update in a couple seconds.

KevinMulhern (approved):
Brilliant work @ZachBaird, this is going to make how to switch much easier to find 🎉

PR #1450, opened 2020-10-27
Update form.scss

DoviMaj:
fix https://github.com/TheOdinProject/theodinproject/issues/1408

.form__section--right-aligned class is obsolete in create form and in edit form @KevinMulhern with those changes.

### DIFF
--- a/app/assets/stylesheets/components/form/form.scss
+++ b/app/assets/stylesheets/components/form/form.scss
@@ -18,6 +18,8 @@
 
     &--bottom {
       margin-bottom: 0;
+      display: block;
+      text-align: center;
     }
   }
 
@@ -111,6 +113,8 @@
   &__toggle-checkbox {
     display: flex;
     align-items: center;
+    justify-content: center;
+    margin-bottom: 20px;
 
 
     @include media-breakpoint-down(sm) {
@@ -154,3 +158,11 @@
     }
   }
 }
+
+@media (max-width: 405px){
+  .form__section.form__section--bottom > .button {   
+    display: block;   
+    margin: auto;
+  }
+}
+


KevinMulhern:
The create form has changed at max width, the make solution public label and checkbox should be on the same row as the submit button like it is in production:
Screenshot 2020-10-28 at 01 23 18

KevinMulhern:
The edit form shouldn't have the make solution public label and checkbox on the row above at max width:

Screenshot 2020-10-28 at 01 27 01

app/assets/stylesheets/components/form/form.scss
```diff
@@ -154,3 +158,11 @@
     }
   }
 }
+
+@media (max-width: 405px){
+  .form__section.form__section--bottom > .button {
```
KevinMulhern, line 163: As discussed in discord we should have a wrapper div around the buttons instead.

KevinMulhern (changes requested):

DoviMaj:
@KevinMulhern the screenshots are at which width? when u say max-width what px is that? 

also there where some responsive issues with the create form that i tried resolving at once with the edits. 

submit

KevinMulhern:
Those screen shots are over 1500px

DoviMaj:
ok ive improved the solution to solve both with media queries.

```
@media (max-width: 1510px){
    .form__section--bottom{
    display:block;
    text-align: center;
}
}

@media (max-width: 1153px){
    .form__section--bottom{
    display:flex;
    text-align: center;
}
}

@media (max-width: 652px){
    .form__section--bottom{
    display:block;
    text-align: center;
}
}

.form__section--right-aligned {
    align-items: center;
    justify-content:space-between
}

.form__toggle-checkbox {
    display: flex;
    align-items: center;
    justify-content: center;
    margin-bottom: 20px
}

@media (max-width: 405px){
    .form__section.form__section--bottom > .button {   
    display: block;   
    margin: auto;
    }
}

```

the gifs start at 1800px:

create-solution
edit-solution

KevinMulhern:
Yeah that looks like the correct behaviour @DoviMaj.

Are you testing these changes on a local version of the app? I noticed your using a patch branch.

DoviMaj:
i have a patch branch but im actually just testing inside a inspector stylesheet with dev tools.

should i proceed with those changes @KevinMulhern?

leosoaivan:
@KevinMulhern @DoviMaj could we just update the `react-modal__body` to have a default width of 60%, instead of 40%? The media queries are also included there, so we shouldn't need to include more media query rules. See here: https://github.com/TheOdinProject/theodinproject/blob/master/app/assets/stylesheets/components/react_modal.scss

DoviMaj:
@leosoaivan we can let it 60% by default but that doesn't make it responsive and the overflowing issue persist

leosoaivan:
I think it's possible.

ezgif com-optimize

At the end of the day, I don't think we want to add custom media-query breakpoints, as we use the ones provided by Bootstrap, and changing the default width to 60% is a one-line change. If we add your wrapper div around the buttons as well, I think we have a good solution.

DoviMaj:
If that is the behavior we want il go for it. just pointing out that the content is right aligned on bigger screens and on smaller screens its left aligned and on mobile screens its all messed up. the same goes for the create form.

Also 60 % width looks weird on normal desktop imo, its huge.

Also i did what u said and didnt get the result u got at the screenshot. did u add or chenge some css @leosoaivan ?

leosoaivan:
@DoviMaj 

Yes, I changed the css in the file I linked above. Is that where you tried making the changes as well? We do want the changes for the content as well, of course, but I didn't focus on that in my example.

Like I mentioned above, if the modal is too big in wider screens, could we use the Bootstrap queries to add additional rules?

DoviMaj:
```
/*changed justify content to center and added flex wrap,
would make sense to rename class to center align*/
.form__section--right-aligned {
  justify-content: center;
  flex-wrap: wrap;
}

/*changed text align to center on existing media query*/
@media (max-width: 767.98px) {
.form__section--right-aligned {
  text-align: center;
  }
}

/*added margin and justify content center*/
.form__toggle-checkbox {
  margin-bottom: 10px;
  justify-content: center;
}

/*buttons wrapper div on edit form style*/
.form__section--bottom__buttons-wrapper {
  display: flex;
  justify-content: center;
  align-items: center;
  flex-wrap: wrap;
  gap:10px
}
```
we can change the width to whatever we want with this solution because of the buttons wrapper div and `flex-wrap` magic. if it fits beside the toggle it stays there. if not, it goes down. its fully responsive yey!

What about this one @leosoaivan @KevinMulhern ?

sl2
sl3

leosoaivan:
@DoviMaj
I like this solution a lot better. And you are right `.form__section--right-aligned` doesn't make sense anymore. What if we combined all the rules under `.form__section--bottom` and get rid of all instances of `.form__section--right-aligned` in the code? What do you think about that @KevinMulhern?

Also, for that media query, I think we can use the Bootstrap mixin `@include media-breakpoint-down(sm) { ... }`.

DoviMaj:
About the media query i didnt add it. i used the existing one that i belive is called: `@include media-breakpoint-down(md)` its just showing like that cause im using dev tools to play with it.

DoviMaj:
Whats up with this @KevinMulhern?

KevinMulhern:
Thanks for all your work on this @DoviMaj 💪

DoviMaj:
should i proceed with this changes? @KevinMulhern

DoviMaj:
@KevinMulhern did u open a new issue?

KevinMulhern:
I didn't, I'll reopen the old one.

DoviMaj:
ok. I'm just waiting for aproval of the latest solution and the changes to the class names @KevinMulhern whenever u can take a look.

KevinMulhern:
> ok. I'm just waiting for aproval of the latest solution and the changes to the class names @KevinMulhern whenever u can take a look.

Where is the latest solution at @DoviMaj?

DoviMaj:
> ```
> /*changed justify content to center and added flex wrap,
> would make sense to rename class to center align*/
> .form__section--right-aligned {
>   justify-content: center;
>   flex-wrap: wrap;
> }
> 
> /*changed text align to center on existing media query*/
> @media (max-width: 767.98px) {
> .form__section--right-aligned {
>   text-align: center;
>   }
> }
> 
> /*added margin and justify content center*/
> .form__toggle-checkbox {
>   margin-bottom: 10px;
>   justify-content: center;
> }
> 
> /*buttons wrapper div on edit form style*/
> .form__section--bottom__buttons-wrapper {
>   display: flex;
>   justify-content: center;
>   align-items: center;
>   flex-wrap: wrap;
>   gap:10px
> }
> ```
> 
> we can change the width to whatever we want with this solution because of the buttons wrapper div and `flex-wrap` magic. if it fits beside the toggle it stays there. if not, it goes down. its fully responsive yey!
> 
> What about this one @leosoaivan @KevinMulhern ?
> 
> sl2
> sl3

DoviMaj:
after that Leo also had some coments: 

> @DoviMaj
> I like this solution a lot better. And you are right `.form__section--right-aligned` doesn't make sense anymore. What if we combined all the rules under `.form__section--bottom` and get rid of all instances of `.form__section--right-aligned` in the code? What do you think about that @KevinMulhern?
> 
> Also, for that media query, I think we can use the Bootstrap mixin `@include media-breakpoint-down(sm) { ... }`.

PR #1451, opened 2020-10-27
Added rounded corners to modal

DoviMaj:
To match TOP design


leosoaivan (approved):
The modal is indeed rounded.

PR #1482, opened 2020-11-11
Adding daily reporting for project submission votes.

ZachBaird:
This code change:

- Adds a Vote model class we can use for things like the created_at scope which
- Was added to the daily report.

Additionally, repetitive "today"'s were removed at @KevinMulhern 's request.


KevinMulhern (approved):
🐧

Issue #1480, opened 2020-11-11
Flip Async and React section in seeds file.


I3uckwheat:
![image](https://user-images.githubusercontent.com/4411701/98856710-d36bca00-2412-11eb-8e52-7870baaf6e31.png)
These two sections should be flipped so we're not making mental jumps between learning vanilla JS and a framework. In the future, this could mean we need to add an API project to the React stuff as well.

twalton83:
I’ll go ahead and handle this.

PR #1483, opened 2020-11-11
Swapped order of React and Async lessons

twalton83:
Resolving Issue #1480

- Swapped order of React and Async/API lessons so Async/API now comes before React in the Javascript course.


KevinMulhern (approved):
Nice work 👍

CouchofTomato:
Thanks @twalton83

PR #1488, opened 2020-11-14
Add docker compose

inductor:
A follow-up pull request to https://github.com/TheOdinProject/theodinproject/pull/1250

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,6 @@ ENV RAILS_ENV=docker
 COPY . .
 
 EXPOSE 3000
-CMD ["entrypoint.sh"]
+CMD ["/rails/entrypoint.sh"]
 
 
--- a/config/database.yml
+++ b/config/database.yml
@@ -10,7 +10,7 @@ default: &default
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   timeout: 5000
 
-development:
+development: &development
   <<: *default
   database: <%= ENV['POSTGRES_DEV_DB'] || 'theodinproject' %>
 
--- a/config/database.yml
+++ b/config/database.yml
@@ -22,9 +22,8 @@ test: &test
   database: <%= ENV['POSTGRES_TEST_DB'] || 'theodinproject_test' %>
 
 docker:
-  <<: *default
+  <<: *development
   database: <%= ENV['POSTGRES_DEV_DB'] || 'theodinproject' %>
-  username: <%= ENV['POSTGRES_USERNAME'] || 'postgres' %>
 
 production:
   adapter: postgresql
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -19,6 +19,6 @@ ENV RAILS_ENV=docker
 COPY . .
 
 EXPOSE 3000
-CMD ["entrypoint.sh"]
+CMD ["/rails/entrypoint.sh"]
```
mindovermiles262, line 22: Since we set `WORKDIR /rails` previous in the Dockerfile, we don't _need_ the full path but it really doesn't matter. No need to fix

config/database.yml
```diff
@@ -22,9 +22,8 @@ test: &test
   database: <%= ENV['POSTGRES_TEST_DB'] || 'theodinproject_test' %>
 
 docker:
-  <<: *default
+  <<: *development
```
mindovermiles262, line 25: If we change this to `*deveopment` I don't think we need the following `database: ... ` line as it's  a duplicate of Development

docker-compose.yaml
```diff
@@ -0,0 +1,26 @@
...
+    depends_on:
+      - db
+    environment: 
+      POSTGRES_USERNAME: postgres
+      POSTGRES_PASSWORD: theodinproject
+      POSTGRES_DEV_DB: theodinproject
+      POSTGRES_HOST: db
+
+volumes:
+  postgres-data:
```
mindovermiles262, line 26: This all looks great! 👍

mindovermiles262 (commented):
Also seeing:
```
```

Dockerfile
inductor: it is necessary. or, it's has to be either in the `$PATH` or `./entrypoint.sh`

config/database.yml
inductor: It didn't work without it

inductor:
@mindovermiles262 The most interesting thing is that the current dockerized environment does not already work well. I just fixed it, but I get migration error.

mindovermiles262:
Same with me RE: migrations not working. I'll poke at the migrations to see what I can do

mindovermiles262:
Merged - 

I merged this PR and quickly added the fixes. See #1490 if interested.

Basically I removed the 'docker' rails environment since we dont' really need it and made docker run off the development environment.

We needed to use `rails db:setup` instead of `rails db:create` and `rails db:migrate`. The difference is subtle but the `db:setup` command will build the DB from the schema and not the migration files.

Also needed to add some `yarn install` commands to the docker container so that they would compile.

Everything seems to be working now, and we all really appreciate your contribution @inductor 👍  Thank you so much!

PR #1491, opened 2020-11-16
Added additional sort to project submissions.

ZachBaird:
This code:

- Adds an additional sort to the ProjectSubmission, enabling it to sort by most likes and then by most recent date.


CouchofTomato (approved):
Thanks @ZachBaird

PR #1496, opened 2020-11-16
Fixing dupe submission bug and adding likes to current_user's submission.

ZachBaird:
This code: 

- Adds liking functionality to the current user's submission, allowing them to like their own project submission.
- Removes prepending the list of submissions with the user's submission.
- Renders the user's submission separate from the rest so that user submission renders at the top and isn't duped when visibility is changed.


ZachBaird:
This pull request will seem some updates tomorrow, please do not merge yet.

PR #1534, opened 2020-11-26
Updated Ruby path description

kadenjt:
Wording no longer made sense with the new curriculum structure.


marvingay (approved):
thank you for your contribution.

PR #1539, opened 2020-11-29
Add enough spacing for usernames

tmannazy:
Fixed the spacing to accommodate long usernames on the screen size of 770px - 1000pxon the Submission page https://www.theodinproject.com/courses/javascript/lessons/tic-tac-toe-javascript.


.ruby-version
```diff
@@ -1 +1 @@
-2.6.6
+2.7.0
```
I3uckwheat, line 1: I don't think we want to have these changes in the project, this is going to cause issues with the deployment. We need to stay on the same version as before.

Gemfile.lock
```diff
@@ -82,7 +82,7 @@ GEM
       activesupport (>= 3.0.0, < 6.1)
       ruby2_keywords (>= 0.0.2, < 1.0)
     ast (2.4.1)
-    autoprefixer-rails (10.0.1.3)
+    autoprefixer-rails (10.0.3.0)
```
I3uckwheat, line 85: I don't think we want to modify the .lock file for the same reason as above. It could cause issues with the deployment that's setup to run with Ruby 2.6.6

app/assets/stylesheets/components/card.scss
```diff
@@ -84,4 +84,4 @@
   &__title {
     margin-bottom: 60px;
   }
-}
+}
```
I3uckwheat, line 87: Please add a newline at the end of this file

app/assets/stylesheets/components/project_submissions/submissions.scss
```diff
@@ -140,4 +143,4 @@
     padding: 1.5em 0em 0.5em;
     margin-bottom: 0;
   }
-}
+}
```
I3uckwheat, line 146: Please add a newline at the end of this file

app/assets/stylesheets/components/project_submissions/submissions.scss
```diff
@@ -1,6 +1,7 @@
 .submissions {
   margin-bottom: 2em;
   text-align: left;
+  width: 650px;
```
I3uckwheat, line 4: Setting the width here is not something we can do. After growing the screen, we see this: 
image

This will have to continue to be responsive.

I3uckwheat (changes requested):
Be sure to test all sizes when making CSS changes like this, it's easy to forget to remember growing the screen after shrinking it and making changes.

I3uckwheat:
This solution also does not handle even longer usernames than what was tested. Consider truncating the username or finding another solution to handle this.

KevinMulhern:
Sorry, this has been fixed in another PR. Thanks for contributing.

Issue #1507, opened 2020-11-19
[Tests - JS/React] Add unit tests for modal.js


leosoaivan:
### Background:
The recent addition of the project submissions feature introduces React components to The Odin Project's website. In order to maintain a level of confidence on the feature's implementation, we would like to add tests, starting with unit tests.

For this issue, add unit tests for `component/modal.js`.

### Suggestions:
1. Navigate to the TOP website, identify the component in need of testing.
2. Using `react-testing-library`, write tests to address the side effects of changing props/state. Don't test props/state themselves. For more information, Google "white-box testing".

### Acceptance Criteria:
1. Running `yarn run test` does not break the current testing suite.

aayushmau5:
Hey, i wanna take this one.

leosoaivan:
Resolved by https://github.com/TheOdinProject/theodinproject/pull/1549.

PR #1549, opened 2020-11-30
Add unit tests for modal.js

aayushmau5:
#1507 

This commit:
- Adds unit tests for the modal.js component
- Adds aria labels in component for accessibility purposes.

### DIFF
--- a/app/javascript/components/project-submissions/components/__tests__/modal.test.js
+++ b/app/javascript/components/project-submissions/components/__tests__/modal.test.js
@@ -3,7 +3,7 @@ import React from "react";
 import Modal from "../modal";
 
 describe("Modal", () => {
-  test("When the modal is not shown", () => {
+  test("Checks for the specified classname on modal when the modal is not shown", () => {
     const { getByLabelText } = render(
       <Modal 
         handleClose={() => {}} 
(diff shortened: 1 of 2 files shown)


app/javascript/components/project-submissions/components/__tests__/modal.test.js
```diff
@@ -0,0 +1,48 @@
...
+    const { getByLabelText } = render(
+      <Modal 
+        handleClose={() => {}} 
+        show={false}
+      >
+        {" "}
+      </Modal>
+    );
+
+    expect(getByLabelText("modal--hidden")).toHaveClass("react-modal--hidden");
```
leosoaivan, line 16: Test should not rely solely on an assertion on a class name, which is opaque and unnecessary to a user. Add an assertion that tests for the presence/absence of children, per line 14 of the component. You may want to pass in a simple `div` with some text to assert against.

app/javascript/components/project-submissions/components/__tests__/modal.test.js
```diff
@@ -0,0 +1,48 @@
...
+    const { getByLabelText } = render(
+      <Modal 
+        handleClose={() => {}} 
+        show={true}
+      >
+        {" "}
+      </Modal>
+    );
+
+    expect(getByLabelText("modal--shown")).toHaveClass("react-modal");
```
leosoaivan, line 29: Same here: Add an assertion that tests for the presence/absence of children, per line 14 of the component.

leosoaivan (changes requested):
Excellent start.

aayushmau5:
@leosoaivan Please check now.

leosoaivan (approved):

PR #1555, opened 2020-12-01
Add max width to submission left container to prevent overflowing

sher-s7:
**Before**: 
before

**After**:
after


I3uckwheat:
Hey @sher-s7 , this was assigned to someone else, and this was *also* fixed here: https://github.com/TheOdinProject/theodinproject/pull/1537

Make sure you are communicating on the issues so we don't redo work.

PR #1580, opened 2020-12-11
Update 05_html_css_course_seeds.rb

DoviMaj:
Changed HTML and CSS course description. since it doesnt make much sense for those not comming from the ruby path.


marvingay (approved):
Thank you for your contribution.

PR #1582, opened 2020-12-13
Added missing spaces

zainiscoding:
There were some missing spaces, so for example, the text read:
Showing10 most recent submissions - View full list of solutions orView old submissions

If I'm not mistaken, my changes should address this.


KevinMulhern (approved):
Thanks @zainiscoding, looks great 💯

PR #1597, opened 2020-12-19
Fix/lesson content center align

DoviMaj:
@KevinMulhern please let me know if this is acceptable

# Screenshoots
# previous with nav
Screenshot_2020-12-19 How Does the Web Work The Odin Project

# current with nav
Screenshot_2020-12-19 How Does the Web Work The Odin Project(1)

# previous without nav
Screenshot_2020-12-19 Setting Up Git The Odin Project

# current without nav
Screenshot_2020-12-19 Setting Up Git The Odin Project(1)

### DIFF
--- a/app/assets/stylesheets/components/lesson/_navigation.scss
+++ b/app/assets/stylesheets/components/lesson/_navigation.scss
@@ -1,3 +1,7 @@
+.lesson-nav-wrapper {
+  margin-right: -60px;
+}
+
 .lesson-navigation {
   padding-bottom: 82px;
   position: -webkit-sticky;
@@ -32,7 +36,6 @@
   }
 
   &__link {
-
     &.active {
       color: $black;
       font-weight: bold;
--- a/app/assets/stylesheets/components/lesson/_navigation.scss
+++ b/app/assets/stylesheets/components/lesson/_navigation.scss
@@ -1,5 +1,6 @@
 .lesson-nav-wrapper {
-  margin-right: -60px;
+  display: flex;
+  justify-content: center;
 }
 
 .lesson-navigation {
--- a/app/assets/stylesheets/components/lesson/lesson.scss
+++ b/app/assets/stylesheets/components/lesson/lesson.scss
@@ -1,0 +1,3 @@
+.lesson {
+  max-width: 1240px;
+}
--- a/app/assets/stylesheets/components/lesson/lesson.scss
+++ b/app/assets/stylesheets/components/lesson/lesson.scss
@@ -1,3 +1,5 @@
 .lesson {
-  max-width: 1240px;
+  @include media-breakpoint-up(lg) {
+    max-width: 1240px;
+  }
 }
--- a/app/assets/stylesheets/components/lesson/lesson.scss
+++ b/app/assets/stylesheets/components/lesson/lesson.scss
@@ -1,5 +1,0 @@
-.lesson {
-  @include media-breakpoint-up(lg) {
-    max-width: 1240px;
-  }
-}
--- a/app/assets/stylesheets/components/lesson/lesson_container.scss
+++ b/app/assets/stylesheets/components/lesson/lesson_container.scss
@@ -1,0 +1,5 @@
+.lesson-container {
+  @include media-breakpoint-up(lg) {
+    max-width: 1710px;
+  }
+}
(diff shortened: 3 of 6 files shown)


KevinMulhern:
Hey @DoviMaj, sorry this one has been sitting around for a while. I was having a look through the bootstrap docs and noticed they have a `container-fluid`. I wonder if using that would be the more standard™️ way of solving this issue.

I'm a little bit nervous about using negative margins, it feels a bit too brittle.    

https://getbootstrap.com/docs/4.0/layout/overview/

DoviMaj:
hey @KevinMulhern. no prob. i see, i tried it with ` container-fluid ` and it solves the width problem but its not responsive, on all viewports it will take 100%. this gif is with `container-fluid`

Peek 2020-12-29 01-36

also the nav will be pretty much on the end of the viewport which dosent look so nice and still taking up a lot of white space on the right.

chrome-capture (4)

but maybe u meant something else?

KevinMulhern:
Yeah I think you will have to experiment with the col sizes for the larger widths. Bootstrap has different classes for different break points you can try playing with https://getbootstrap.com/docs/4.0/layout/grid/#grid-options

DoviMaj:
current behavior
current
</br>
with different col sizes and making the nav container `display: flex` and `justify-content: center` the best i can get is this the screenshot bellow, but notice that the lesson content isnt 100% aligned to the center and on screens larger then  1700px the lesson content gets huge because of `container-fluid`
Peek 2020-12-29 21-02

KevinMulhern:
I think its not centering because the the lesson content col is too large, There are 12 possible columns and the side nav is 3 so to center the lesson content its col needs to be a 6 to leave 3 on the other side. 

try changing the lesson content col to .col-xl-6 and .col-lg-6, it might not even need the xl col class.

DoviMaj:
i tried with `col-lg-6` and it does center the content but still its huge on large screens because its 6 col of 100% width..
and on smaller screens its too small imo.

Peek 2020-12-30 18-38

1024px :
chrome-capture (5)
1440px:
chrome-capture (6)
2260px:
chrome-capture (7)

what do u think @KevinMulhern ?

KevinMulhern:
That looks good to me @DoviMaj, maybe we could give the `lesson-content` a max width for larger screens. But otherwise I think this is the way to go 🚀

DoviMaj:
Do u mean `.lesson` ? because changing `lesson-content`'s `max-width` will make it not be centred. 
changing `lesson` is what i had done in the first solution attempt, which is basically overwriting bootstrap's `container-fluid` for larger screens.  
`.lesson {
    max-width: 1710px;
}`
makes `lesson-content` have a `max-width` of `825px` the exact amount it has currently on larger screens.

Also isnt it too small on smaller screens? 
- 1024px 
changed:
`lesson-content` has 482px width
chrome-capture (9)
current:
`lesson-content` has 690px width
chrome-capture (10)

KevinMulhern:
No not the `.lesson` class, it wraps everything. But we only want a max-width on the lesson content for larger screens.

Can you update this PR with your latest changes please? it would be easier for me to have a play with it and get back to you on what else needs done.

DoviMaj:
updated the PR @KevinMulhern

KevinMulhern:
Just had a play with it @DoviMaj, its looking really good.

Just a couple of tweaks and I think this will be good to go.

1. Can you add `max-width: 825px;` and   `margin: 0 auto;` to the `.lesson-content` class please? that will keep the lesson content the same width as it is now on extra large screens.

2. The transition from extra large to large screens could be a bit smoother, `col-xl-6 col-lg-8` on that column seems to work nicely.

DoviMaj:
Sure. that seems good!

My only concern is the nav bar that is extremly to the left on xxl devices.

chrome-capture (13)

if we shouldnt use `.lesson` we can add a `.lesson-container` class and have `max-width` there without needing to add `max-width` to `lesson-content`.
```
.lesson-container {
    max-width: 1710px;
}
``` 
gives 825px to `lesson-content` and has the position of the nav the same as current.
chrome-capture (14)

KevinMulhern:
If that still works with the sticky scroll then lets get this over the line and go with that.

DoviMaj:
dont merge yet..i have to change something still

DoviMaj:
Ok. this is good to go imo. Let me know if its all good.

KevinMulhern (approved):
Works brilliantly @DoviMaj, thanks for all your work on this.

PR #1611, opened 2020-12-30
Fix casing of JavaScript

karlhorky:
Hi there, thanks for The Odin Project - really cool! I think I'll be contributing more often :)

Not sure if I'm doing this correctly, but I wanted to fix the casing of JavaScript in the banner (and found a few more occurrences where it was incorrect):

Screen Shot 2020-12-30 at 11 12 06

I first looked in [TheOdinProject/curriculum](https://github.com/TheOdinProject/curriculum), but that doesn't seem to be the correct place for it... somewhat confusing, where to find the information here.


karlhorky:
Thanks for the merge!

PR #1695, opened 2021-02-05
Updated Node.js card and Express card title to include MongoDB

CatQueenCodes:
This PR is because I have been noticing people in discord frequently questioning why the JS path has no Database section. The Node card description is lacking compared to the rails one. To remedy this I made the Node card more enticing and added mention of MongoDB so when people are browsing the overviews of both paths when deciding they see mention of databases on the Node path as well ass updating the "Express" section to "Express & MongoDB" so it is clear where Databases are introduced.

Screen Shot 2021-02-05 at 1 10 27 PM
Screen Shot 2021-02-05 at 1 10 34 PM
Screen Shot 2021-02-05 at 1 17 35 PM

### DIFF
--- a/db/seeds/08_node_js_course_seeds.rb
+++ b/db/seeds/08_node_js_course_seeds.rb
@@ -77,7 +77,7 @@ section = create_or_update_section(
   title_url: 'Express'.parameterize,
   course_id: course.id,
   position: section_position,
-  description: 'Here we finally get to Express, the most popular back-end JavaScript framework.'
+  description: 'Here we finally get to Express, the most popular back-end JavaScript framework, and MongoDB, a non-relational database frequently paired with Node.',
 )
 
 lesson_position += 1
--- a/db/seeds/08_node_js_course_seeds.rb
+++ b/db/seeds/08_node_js_course_seeds.rb
@@ -77,7 +77,7 @@ section = create_or_update_section(
   title_url: 'Express'.parameterize,
   course_id: course.id,
   position: section_position,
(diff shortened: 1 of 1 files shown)


db/seeds/08_node_js_course_seeds.rb
```diff
@@ -73,7 +73,7 @@
 
 section_position += 1
 section = create_or_update_section(
-  title: 'Express',
+  title: 'Express & MongoDB',
   title_url: 'Express'.parameterize,
   course_id: course.id,
   position: section_position,
```
twalton83, line 79: Can we also update the description since it no longer only indicates that it houses the Express content?

twalton83 (commented):
Just added a note about also updating the description of the Express card.

db/seeds/08_node_js_course_seeds.rb
CatQueenCodes: I was thinking about that as well. I have not used Express or MongoDB so not sure if the below is good phrasing

How does "Here we finally get to learning the Express framework and integrating MongoDB" sound?

Old: Here we finally get to Express, the most popular back-end JavaScript framework.

twalton83: What about the below: 
> Here we finally get to Express, the most popular back-end JavaScript framework, and MongoDB, a non-relational database frequently paired with Express. 

How do you feel about that?

CatQueenCodes: Sounds good to me, I'll update it. Thanks :)

CatQueenCodes:
All descriptions updated.

marvingay (approved):

twalton83 (approved):
Thanks for taking care of this :)

Issue #1666, opened 2021-01-20
Auto-resizing of student project submissions overflow is off. 


CatQueenCodes:
![Screen Shot 2021-01-20 at 2 20 08 PM](https://user-images.githubusercontent.com/59713017/105224180-d974da00-5b2a-11eb-8452-635e9c451cce.png)

It seems to happen when I have it around half screen window size, if I keep making the screen smaller eventually it does wrap around. It only happens with the person who submitted it's line since it has more content than the other lines. 

Not a major issue, just something I noticed

aayushmau5:
@CatQueenCodes I would like to work on this issue.

aayushmau5:
![demo](https://user-images.githubusercontent.com/54525741/105341964-bcf9a000-5c05-11eb-8d96-e8808a41bcde.png)

Is this better?

CatQueenCodes:
We think it looks a bit odd having it as the only item needing 2lines/ having the eye on a seperate line. Would it be possible to try and have it so the buttons+icon shrink to fit

aayushmau5:
Working on it.

aayushmau5:
Here are some of the things i tried:

After setting the font size to 15px.
![first](https://user-images.githubusercontent.com/54525741/105385568-83428c80-5c39-11eb-9684-586793ea261a.png)
After adding `work-break`
![second](https://user-images.githubusercontent.com/54525741/105385575-850c5000-5c39-11eb-8d4b-2ab6b4606bc3.png)

rankoliang:
![image](https://user-images.githubusercontent.com/37024941/105418112-9853f700-5c0a-11eb-842c-33d3ce11fef3.png)

Here's a solution I came up with. Also added color to the borders to emphasize that it's your own submission. I'm not sure how it would look when it comes to the project submissions in the dashboard though.

DoviMaj:
`*, *::before, *::after {
    box-sizing: border-box; 
 }` from bootstrap is causing this. how to fix it? no idea

DoviMaj:
Can someone assign this to me? @aayushmau5 let me take it

DoviMaj:
May be closed

PR #1737, opened 2021-03-02
Fix/solutions overflow

DoviMaj:
Solution for https://github.com/TheOdinProject/theodinproject/issues/1666.

Changes: 
* Edit icon instead of the button
* Link to user dashboard on own solution
* Word break for long names on smaller devices
* Flex gap for submission buttons instead of margin-left (because: on mobile the edit icon didn't have proper margin)

Before:
chrome-capture (48)
After: 
chrome-capture (49)

I took a look at the tests and they are passing but the new case of `SubmissionTitle` returning a link or not depending on `isCurrentUsersSubmission`  isn't being tested. not sure if it needs tests though.


KevinMulhern (changes requested):
This is looking really good @DoviMaj, thanks for taking the time to do this.

I think we should replace the toggle visibility icon with the edit icon however. It looks a bit odd to have an 
icon followed by two buttons and then another icon, replacing the toggle visibility icon aligns the actions and makes them more uniform.

Screenshot 2021-03-02 at 18 32 30

DoviMaj:
Let me change the edit icon width to match the flag..

DoviMaj:
Latest change isnt directly related to the issue at hand. It changes the responsive behavior of the buttons. Because they where taking unnecessary space vertically and didn't align nicely on small to medium viewports.

How it looks now:
Peek 2021-03-03 16-37

KevinMulhern:
Nice work @DoviMaj 💪  this is looking great!

Would you mind removing the VisibleToggle component please? I don't think theres any reason to keep it around anymore since your refactor negates the need for it 😎

DoviMaj:
@KevinMulhern, I'm not sure if u get notifications for open pull requests commits so I'm pinging u here to let u know I have removed the visible toggle component and associated styles as requested, so u can review when u can/want. :smiley: 

Also to ask about the need for tests for this refactor.

KevinMulhern:
Thanks for removing that component @DoviMaj 🎉 

We could add a test for asserting the users title is a link in the submission-title component tests. I think thats the only test this needs. Whats your thoughts?

DoviMaj:
Maybe test also if it's rendering correctly depending on if it is the current user's submission or not

```
  it('displays username if not in dashboard view', () => {
    render(<SubmissionTitle submission={submission} />);
    expect(screen.getByText('TestUser')).toBeInTheDocument();
  });
  it('displays username inside p tag if not in dashboard view and isnt current users submission', () => {
    //not sure why isn't working, returns DIV instead of P
    render(<SubmissionTitle submission={submission} />);
    expect(screen.getByText('TestUser').tagName).toBe('P');
  });
  it('displays a link to user dashboard if not in dashboard view and is current users submission', () => {
    render(<SubmissionTitle isCurrentUsersSubmission submission={submission} />);
    expect(screen.getByText('TestUser')).toHaveAttribute('href', '/dashboard');
  });
```

KevinMulhern:
I don't think we should assert on specific tags like `p`. It makes the test very brittle if we update the type of tag we use for it. 

What about doing something like this @DoviMaj?

```
describe('Submission title', () => {

  describe('when on the dashboard view', () => {
    it('displays the lesson title', () => {
      render(<SubmissionTitle submission={submission} isDashboardView />);
      expect(screen.getByText('Test Title')).toBeInTheDocument();
    });

    it('displays a link to the project', () => {
      render(<SubmissionTitle submission={submission} isDashboardView />);
      expect(screen.getByText('Test Title')).toHaveAttribute('href', 'test_path');
    });
  })

  describe('when not on the dashboard view', () => {
    it('displays submissions username', () => {
      render(<SubmissionTitle submission={submission} />);
      expect(screen.getByText('TestUser')).toBeInTheDocument();
      expect(screen.getByText('TestUser')).not.toHaveAttribute('href', '/dashboard');
    });

    it('displays link to users dashboard when submission belongs to the current user', () => {
      render(<SubmissionTitle submission={submission} isCurrentUsersSubmission />);
      expect(screen.getByText('TestUser')).toHaveAttribute('href', '/dashboard');
    });
  })
});
```

DoviMaj:
Yeah, that makes more sense. submited

KevinMulhern:
Looks like this is good to go 🚀 

Thanks for all your work on this @DoviMaj 😄

KevinMulhern (approved):

PR #1821, opened 2021-04-01
Fix: Navbar alignment and position

DoviMaj:
Changes:
* Align Navbar's divider to the center 
* Refactor Navbar's position code

Before:
before

After:
after

### DIFF
--- a/app/assets/stylesheets/components/navbar.scss
+++ b/app/assets/stylesheets/components/navbar.scss
@@ -2,7 +2,7 @@
   max-height: 100%;
   margin-bottom: 50px;
   display: flex;
-  padding: 0 168px;
+  padding: 7px 168px;
   background: #ffffff;
   z-index: 100;
 
@@ -77,8 +77,8 @@
 
     .divider {
       font-size: 2em;
-      line-height: 78px;
       font-weight: 200;
+      padding: 0 0.7rem;
     }
 
     .avatar {
@@ -93,7 +93,8 @@
 
 @media (min-width: 768px) and (max-width: 1024px)  {
   .navbar{
-    padding: 0px 30px 0px 30px;
+    padding: 12px 30px 0px 30px;
+    margin-bottom: 62px;
   }
 }
 
--- a/app/assets/stylesheets/components/navbar.scss
+++ b/app/assets/stylesheets/components/navbar.scss
@@ -76,9 +76,13 @@
     }
 
     .divider {
-      font-size: 2em;
-      line-height: 78px;
-      font-weight: 200;
+      line-height: 76px;
+
+      &:after{
+        font-size: 1.5em;
+        content: "|";
+        padding: 0;
+      }
     }
 
     .avatar {
--- a/app/assets/stylesheets/components/navbar.scss
+++ b/app/assets/stylesheets/components/navbar.scss
@@ -76,8 +76,9 @@
     }
 
     .divider {
-      font-size: 1.5em;
+      font-size: 2em;
       line-height: 78px;
+      font-weight: 200;
     }
 
     .avatar {
--- a/app/assets/stylesheets/components/navbar.scss
+++ b/app/assets/stylesheets/components/navbar.scss
@@ -2,7 +2,7 @@
   max-height: 100%;
   margin-bottom: 50px;
   display: flex;
-  padding: 7px 168px;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/navbar.scss
```diff
@@ -92,7 +93,8 @@
 
 @media (min-width: 768px) and (max-width: 1024px)  {
   .navbar{
-    padding: 0px 30px 0px 30px;
+    padding: 12px 30px 0px 30px;
+    margin-bottom: 62px;
```
KevinMulhern, line 97: I'm not sure if there is a good enough reason to include these styles? 

The logo should sit snug to the top of the screen as designed the 50px of margin the navbar has by default is that size to fit the visual hierarchy of the spacing.

app/views/shared/_navbar.html.erb
```diff
@@ -26,7 +26,7 @@
           <%= link_to "Community", chat_link, class: 'nav-link', target: '_blank', rel: 'noreferrer' %>
         </li>
 
-        <li class="divider nav-link">|</li>
+        <li class="divider nav-link">l</li>
```
KevinMulhern, line 29: having the letter l here doesn't seem right,  is there a good reason for not using the pipe that was here before?

app/assets/stylesheets/components/navbar.scss
```diff
@@ -2,7 +2,7 @@
   max-height: 100%;
   margin-bottom: 50px;
   display: flex;
-  padding: 0 168px;
+  padding: 7px 168px;
```
KevinMulhern, line 5: I can see the reason for including this, but the navbar was designed to be in that position at the top of the screen so I think we should leave this out of the PR.

KevinMulhern (commented):
Hey @DoviMaj, thanks for taking the time to do this. It will be great to fix that alignment issue.

However I don't think we should be doing anything with the positioning, as there isn't currently a problem with it and it's very subjective if these changes improve it or not.

app/views/shared/_navbar.html.erb
DoviMaj: The pipe won't align. I tried my best lol.

app/assets/stylesheets/components/navbar.scss
DoviMaj: This was needed for some obscure reason to match the previous position when the divider was ruling. But not needed if we revert that

app/assets/stylesheets/components/navbar.scss
DoviMaj: This is only relevant if remove the divider as the ruler. if we do the navbar sticks to the top of the screen that's why I needed to add some padding to the top to keep it in his original position.

DoviMaj:
@KevinMulhern let me know when u can if I should revert the position changes and what u think about the divider. :peace_symbol:

KevinMulhern:
Hey @DoviMaj, sorry mate I should have been clearer the other day. 

Yeah can we revert the positional changes please and I think we need to find a better way to handle the divider. It could play havoc with screen readers.

DoviMaj:
@KevinMulhern, cool. Just to be sure, u saw this https://github.com/TheOdinProject/theodinproject/pull/1821#discussion_r606022132 right?

KevinMulhern:
Yep I saw that @DoviMaj, imo it would be worth having a look at  how other sites implement this. This _should_ be a solved problem.

DoviMaj:
@KevinMulhern, from my research, the usual way is to use CSS, but if that won't work for us because of the width then..

I've now found two ways to align the pipe. 

1st. duh..

```
    position: relative;
    top: -2.5px;
```

2nd. 

Add a span inside the `li` and `:after` it add 
```    
   content: "|";
   padding: 0;
```

from here: https://css-tricks.com/forums/topic/trouble-with-a-divider-pipe-in-my-navbar/

KevinMulhern:
2nd solution looks like the one to go with for me 👍

DoviMaj:
piedpiper
Note to self: Never underestimate the complexity of aligning a pipe

KevinMulhern:
Looks great to me @DoviMaj, thanks for all the effort you put into this 💪

KevinMulhern (approved):

PR #1843, opened 2021-04-14
Fix: support for ios

DoviMaj:
Fix: Support for *Safari

Replaces all `flexbox-gap` instances with `margin`.

### DIFF
--- a/app/assets/stylesheets/components/form/form.scss
+++ b/app/assets/stylesheets/components/form/form.scss
@@ -6,7 +6,10 @@
     justify-content: center;
     align-items: center;
     flex-wrap: wrap;
-    gap:10px
+
+    :first-child{
+      margin-right: 20px;
+    }
   }
   
   &__section {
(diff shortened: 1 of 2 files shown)


KevinMulhern:
Thanks @DoviMaj, this looks great on larger screens 💪 

It looks like we'll need a breakpoint for putting some margin between the buttons mobile screens however:

Screenshot 2021-04-14 at 23 47 08

DoviMaj:
Good catch! done

DoviMaj:
Actually, I checked and `:nth-child` doesnt have great support for Opera, Samsung and Chrome for Android. Better use a different solution right?

https://caniuse.com/?search=%3Anth-child

KevinMulhern:
I think it should be fine @DoviMaj, the support for the newer version of :nth-child child isn't great but you're not using any of its features here mate 😄

KevinMulhern (approved):
Great work @DoviMaj, looks ace on mobile.

Issue #908, opened 2018-06-11
Design: Dark theme


adachiu:
I've created the dark theme for the lesson page. Since most of the reading is on the lesson pages, I think we can have the dark theme implemented on lesson pages first. The component underneath consisting of "view course" and "next lesson" may or may not need a dark theme. Let's have a discussion around this.

*Edit*
Will have a mock for another lesson page with more page variety. Ie. code blocks, Assignment block, etc.

![dark-lesson-detail-1](https://user-images.githubusercontent.com/13413986/41238182-e6b0981a-6dc7-11e8-94d8-140781c114d4.jpg)

Deetss:
:+1:

CouchofTomato:
@adachiu It looks amazing and will definitely be a nice choice to give users. Thanks for putting this together.

mindovermiles262:
Hey @adachiu this looks really good. Thanks for putting it together :smiley: 

A few suggestions from the peanut gallery over here:

- The top SVG round icon for "ruby on rails" looks very out of place. I don't know much about SVG graphics but if it's simple to replace the colors it may be worth looking into - something closer to the dark gray of the background so it's not so flashy.

- The very light gray background in the student solutions area is a little bright for a dark theme.  Maybe we can darken it a bit to match the dark theme above.

adachiu:
Thanks for the input @mindovermiles262 !

- There are pros and cons for the badge (top SVG icon) to have a different color. The con would be that the dark theme badge may be confusing from the original set of badges. The pro would be the whole page aligning more to the dark theme if the badge was darker. Or we can try to keep the original badge and adjust the top to allow a smooth transition into the dark theme.

- The bottom solutions part has a dark theme in this newest upload 😄 

![dark-lesson-detail-1 2](https://user-images.githubusercontent.com/13413986/41295892-353af50e-6e8e-11e8-9766-aaf968f705e6.jpg)

CouchofTomato:
I think that looks awesome @adachiu 

Thanks for putting this together.

KevinMulhern:
This looks fantastic @adachiu, great work 🎉

kashura:
Wow this would be awesome!

LE-HU:
Guys just to let you know what i've been using so far when coding @ night. These examples might provide you with some example and ideas in further development of in-app dark mode. Both are mozilla add-ons.
[...]

PR #1854, opened 2021-04-17
Dark background add

diegoromerodev:


twalton83 (commented):
I’m not sure this is what we want.

twalton83:
Please feel free to discuss dark mode on the existing issue #908, but overriding a global style isn’t what we’d want here.

PR #1882, opened 2021-04-27
Clicking AllPaths does not select Community

BrianWisniewski:
Fixed issue where clicking "All Paths" causes the "Community" tab to highlight. #1832 

Attached is a photo of the working fix both logged out and logged in. Below is the rspec test in which two failures related to welcome email job occur. I can only assume this is because I did not set up the mail chimp integration. If these failures are a problem I can go back and look for errors but I was getting these before making changes so I don't believe they will be.
loggedOut
LoggedIn
rspec

### DIFF
--- a/app/views/shared/_navbar.html.erb
+++ b/app/views/shared/_navbar.html.erb
@@ -22,7 +22,7 @@
           <%= link_to 'All Paths', paths_url, class: 'nav-link' %>
         </li>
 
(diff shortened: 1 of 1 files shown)


KevinMulhern (approved):
LGTM 👍 

Thanks for taking the time to fix this @BrianWisniewski

PR #1903, opened 2021-05-02
Fix: lack of margin on like button

DoviMaj:
The margin was to the left instead of to the right

before:
chrome-capture (75)
after:
chrome-capture (76)


KevinMulhern (approved):
Thanks for taking the time to fix this @DoviMaj 

Looks great to me 💯

PR #1910, opened 2021-05-05
adds clarifying copy to path descriptions

twalton83:
There were a significant amount of questions about the order courses in a path should be taken. This change aims to improve student clarity and outlines that courses in the paths should be taken in the order that they are displayed.


xandora:
Copy looks good to me. :+1:

CouchofTomato (approved):

PR #1993, opened 2021-06-04
Chore: Upgrade to Ruby 2.7.1

Blakelist7:
Signed-off-by: Kanwal Singh <kanwal.88singh@gmail.com>

#### Checklist
 - [ ] You have included automated tests for your changes where applicable?

Fixes: #1845


Blakelist7:
@KevinMulhern done but still they are failing :(

KevinMulhern:
Hey @Blakelist7, sorry for the delay. It looks like the new image only comes with bundler 2.xx. We should update bundler in a separate PR. But to get this working for now, I think we will need to change [this line](https://github.com/TheOdinProject/theodinproject/blob/1834c383221bc66cc5a85e2c0c710fd3efca3ae1/.circleci/config.yml#L26)  to `gem install bundler:1.17.3`.

Blakelist7:
@KevinMulhern I have updated that to `2.2.19` in `update Gemfile` commit and now it works fine but still `ci/circleci: tests` fail, I think that's because when we run `rspec`, some warnings are thrown.

KevinMulhern:
I think its the image @Blakelist7, it doesn't have the chrome driver needed for the system tests.  Just a sec I'll find an image that works with node and system tests.

KevinMulhern:
This image works @Blakelist7: `circleci/ruby:2.7.1-buster-node-browsers`

Blakelist7:
@KevinMulhern  Done :)

Gemfile.lock
```diff
@@ -619,4 +619,4 @@ RUBY VERSION
    ruby 2.6.6p146
 
 BUNDLED WITH
-   1.17.3
+   2.2.19
```
KevinMulhern, line 622: can we revert this change please? just to keep this PR focused on the Ruby 2.7.1 update and leave updating bundler in a separate PR which we can do after this one goes out :)

KevinMulhern (changes requested):
Nice work @Blakelist7 thats the hard part done 💪  We just need to fix these depreciation warnings before we can merge:
Screenshot 2021-06-04 at 19 38 08


Looks like its just three classes that need some changes and they all need the same changes. If you need any help with them let me know :)

Gemfile.lock
Blakelist7: Sure!

KevinMulhern:
Hey @Blakelist7, I just noticed It's not deploying to the review apps on heroku. But that should be easy to fix.

We just need to update [this line](https://github.com/TheOdinProject/theodinproject/blob/main/Gemfile#L2) to `ruby '2.7.1'`

Blakelist7:
> Nice work @Blakelist7 thats the hard part done muscle We just need to fix these depreciation warnings before we can merge:
> Screenshot 2021-06-04 at 19 41 44
> 
> Looks like its just three classes that need some changes and they all need the same changes. If you need any help with them let me know :)

Hey, @KevinMulhern Could you help me with this. I did some research and found out some called "hash splat" but when I use `def initialize(**admin, **flag:)` `rubocop` throws a bunch of errors.

KevinMulhern:
No worries @Blakelist7.

The initialize methods need to be changed to keyword arguments:
```
def initialize(admin:, flag:)
```

It's the call class methods below the initialize methods that need the hash splats:
```
def self.call(**args)
  new(**args).call
end
```

Blakelist7:
Thanks for the help but now some tests are failing.

KevinMulhern:
Yeah, you'll have to update them to use keyword arguments like you did with the initialize method:

```
subject(:service) { described_class.call(admin: admin, flag: flag) }
```

KevinMulhern:
They will also need updated to use keyword arguments in `admin/flags.rb`

Blakelist7:
Sorry to bother you again but could you tell what do I have to change in `./spec/system/admin/flags_spec.rb` actually I have never used ruby before.

KevinMulhern:
> actually I have never used ruby before.

Why did you pick up an issue for upgrading the Ruby version of a Rails app if you don't know ruby???

Blakelist7:
I was curious how to we upgrade versions in a application so .......
Sorry I didn't know that I would turn out like this 😞 .

KevinMulhern:
Thanks for all your work on this @Blakelist7, what you have here will be enough. 

In the future please tell maintainers up front that you don't know the language when contributing to an OSS project. It will save everybody some time if you do.

PR #2131, opened 2021-08-08
Add notifcation index view

CouchofTomato:
Because: We want users to be able to view their notifications

This commit:
- Remove notification argument to NotifyUser as it's redundant
- Update method to DeadLink url to return correct lesson path
- Add Notification ViewComponent
- Add helper method to render svgs


CouchofTomato:
odin-notifications

codyloyd:
design and styles LGTM.  I'm too far out of the loop with Ruby/Rails to comment on the rest tho.

app/components/notification_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# rubocop:disable Lint/MissingSuper
```
KevinMulhern, line 3: Haven't seen this one before 😄 does it not like the initializer?

spec/components/notification_component_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+           id: 1,
+           title: 'test title',
+           message: 'test message')
+  end
+
+  before do
+    render_inline(described_class.new(notification: notification))
+  end
+
+  it 'renders the component with a link to the notification\'s url' do
```
KevinMulhern, line 15: These tests are 🔥

KevinMulhern (approved):
This looks brilliant @CouchofTomato, awesome work mate!

PR #2146, opened 2021-08-15
Edit My Dashboard text

msespos:
#### Checklist
 - [N/A] You have verified the tests and linters all pass against your changes?
 - [N/A] You have included automated tests for your changes where applicable?


msespos:
Looks like my comment didn't go through above. This change is so that we don't start users off on the dashboard with a suggestion that they can take the courses out of order, as we generally discourage that.

KevinMulhern (approved):
Thanks for taking the time to do this @msespos 💪

Issue #2209, opened 2021-09-09
fix the title of a section.


JuanDCeballos:
### Description
If you are in the block and inline lesson (the last lesson at the CSS Foundations) and you press the button move to the next lesson you are directly move to introduction but, introduction to what?

#### Steps to Reproduce
1. you  must be in the lesson block and inline (the last lesson at the CSS Foundations).
2. You press the button move to the next lesson.
3. You are directly move to a lesson without a descriptive title.
![image](https://user-images.githubusercontent.com/80058433/132751492-9d4fc054-90b3-4fa3-882a-137c712609a0.png)

#### Expected Behaviour
1. Must be in the lesson block and inline (the last lesson at the CSS Foundations).
2. the button move to the next lesson on its title attribute should say: Move on 'Introduction to Flexbox'.
![image](https://user-images.githubusercontent.com/80058433/132751038-0b4a140d-c4a1-43b8-b08f-b3799fbc8a95.png)
3. The lesson's title should say: Introduction to Flexbox.

Janak2401:
I'd like to fix this issue.

kashura:
Fixed and merged, thanks @Janak2401 and @JuanDCeballos

PR #2217, opened 2021-09-12
Fixed the intro title of flexbox section

Janak2401:
Because:
Resolves issue#2209: https://github.com/TheOdinProject/theodinproject/issues/2209

This commit:
- Changed the title "Introduction" of the flexbox section from Foundations lesson to "Introduction to Flexbox".

### DIFF
--- a/db/fixtures/lessons/foundation_lessons.rb
+++ b/db/fixtures/lessons/foundation_lessons.rb
@@ -321,7 +321,7 @@ def foundation_lessons
     },
     'Introduction' => {
       title: 'Introduction to Flexbox',
-      description: "Let's get starting with flexbox",
+      description: "Let's get started with Flexbox",
       is_project: false,
       url: '/foundations/html_css/flexbox/flexbox-intro.md',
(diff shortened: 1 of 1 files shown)


db/fixtures/lessons/foundation_lessons.rb
```diff
@@ -320,7 +320,7 @@ def foundation_lessons
       identifier_uuid: '25874e5c-2485-4e62-aaa5-5d6cf65b6b52',
     },
     'Introduction' => {
-      title: 'Introduction',
+      title: 'Introduction to Flexbox',
       description: "Let's get starting with flexbox",
```
kashura, line 324: Change "Let's get starting with flexbox" to "Let's get started with Flexbox"

kashura (changes requested):
Can you make this change as well?

db/fixtures/lessons/foundation_lessons.rb
Janak2401: Yea sure.

kashura (approved):

PR #2288, opened 2021-10-06
Fix: code tag color inside knowledge check

Maheshkumar-novice:
#### Because:
`<code>` tag inside `.knowledge-check-link` which is an `<a>` tag, overrides the color of `<code>` to `grey` from `pink`. It's because bootstrap's reboot which has a selector `a > code` that changes it.

now

then

#### This commit
So I've added a `.knowledge-check-link > code` selector which gives `pink` color to `<code>` tag. `$pink` variable is declared in bootstrap.

For example you can see the knowledge check section [here](https://www.theodinproject.com/paths/foundations/courses/foundations/lessons/fundamentals-part-1#knowledge-check), `==` like things inside `<code>` tag are in color `grey` instead of `pink`.

#### Checklist
 - [ ] You have verified the tests and linters all pass against your changes?
 - [ ] You have included automated tests for your changes where applicable?

### DIFF
--- a/app/assets/stylesheets/components/lesson/lesson_content.scss
+++ b/app/assets/stylesheets/components/lesson/lesson_content.scss
@@ -149,9 +149,14 @@
   .knowledge-check-link {
     color: inherit;
     text-decoration: inherit;
+    
     &:active, &:focus, &:hover, &:visited {
         color: inherit;
         text-decoration: none;
+    }
+    
+    & > code {
+      color: $pink;
     }
   }
 


CouchofTomato:
> > > So I need to change `color: $pink` to `color: #e83e8c`. right? @CouchofTomato
> 
> > 
> 
> > Yeah but it might be worth creating a variable for this colour in https://github.com/TheOdinProject/theodinproject/blob/main/app/assets/stylesheets/settings.scss and then referencing that. We would also need to update it to the variable name where we set code blocks elsewhere.
> 
> 
> 
> I will look into it. Current `$pink` is defined in the bootstrap linked to the site, it is also `#e83e8c`. That color is used for our `code` tags by default because bootstrap defines `code` tags color to `$pink` by default. I think you already aware of it.
> 
> 
> 
> https://getbootstrap.com/docs/4.0/content/code/
> 
> 
> 
> So still we need to create `$pink` variable in `settings.scss`? 

Hey, sorry for the delay in response. Been battling a cold. After pondering on this I think for now your solution will do so I'll approve as is. We do have some changes around this planned for some point in the future so we can deal with any work around this at that time. Thanks for your contribution.

Maheshkumar-novice:
No problem. Hope you are better now. Thank you.

PR #2768, opened 2022-03-22
downgrade react-tabs

marvingay:
Problem:
https://github.com/TheOdinProject/theodinproject/issues/2717

Due to a bug that was introduced into the `react-tabs` dependency. 

This commit:
- Downgrades React-Tabs, such that the Lesson Previewer works as intended.


KevinMulhern (approved):
🎉 all working again

Thanks for doing this @marvingay 💪

Issue #2647, opened 2022-02-17
Course tile box shadow should not be white for dark mode


thatblindgeye:
#### Description

PR #2619 recently added a box shadow effect on hover to course tiles (when not logged in). While this works fine on light mode, in dark mode the color used for this box shadow is white. Box shadows should not be white in use cases such as this, and should remain as a dark color.

If the intent is to have the box shadow still stand out against the background, then the background being lightened a bit in addition to changing the box shadow color would be a better alternative. Ultimately, in dark mode this effect is no long a box "shadow", but rather a box "light".

#### Acceptance Criteria:

1. Use the same color for the box shadow in dark mode as is used in light mode

TruePadawan:
_Use the same color for the box shadow in dark mode as is used in light mode_

I don't understand. The current color for the darkmode box shadow (which is the same as lightmode' box shadow) shouldn't be changed? Then that means we should be changing the background color instead or what?

thatblindgeye:
@TruePadawan The dark mode box shadow should be the same color as the light mode box shadow (so the same shade of black). I should actually wait to see what @KevinMulhern has to say and whether this is something that should just be fixed in the Tailwind conversion (since this isn't a *major* issue).

jonro2955:
Hi, I'm new at contributing and this seems like something I'd like to take on as an assignment to the USING GIT IN THE REAL WORLD lesson of the JS track. 

So let me see if I understand the goal here. The box-shadow properties in both modes are currently the same at rgba(0, 0, 0, 0.15), so we just want to change the background color in dark mode only to make the box shadow in dark mode stand out more right? I can see that I can simply change the "--dark-bg-color" variable at the top of the app/assets/stylesheets/dark-mode.css file to make this change. If this is correct, please don't hesitate to assign this issue to me. If I'm wrong, please clarify what I need to do. Thank you.

jonro2955:
Hi, I see that I can resolve this by changing the box-shadow color of the `.curriculum-details-tile:hover` selector in our `dark_mode.css` file to the same color of `rgba(0, 0, 0, 0.05)` that is used for the same selector in the light mode scss file `curriculum_details.scss`. May I be assigned this issue?

thatblindgeye:
@jonro2955 you're assigned!

PR #2798, opened 2022-03-29
Feature: changes the box shadow color of path cards in the paths page to the same color as the course tiles in the home page

jonro2955:

  - `Fix` - bug fixes

<hr>

**1. Because:**

Issue [#2647](https://github.com/TheOdinProject/theodinproject/issues/2647) changed the box shadow color of the homepage's course tiles in dark mode to be the same as in light mode, but the [paths page](https://www.theodinproject.com/paths) also has path cards that need to have the same box shadow color for consistency and better visibility.    

**2. This PR:**

- Changes the box-shadow color of path cards in the paths page to be the same as the course tiles in the homepage to make all hover effects look consistent and improve visibility of elements.


KevinMulhern (commented):
Hey @jonro2955 thanks for taking the time to do this.  But a 50% shadow on the card component isn't what we want to go with for the overall design unfortunately. 

The current shadows we use here have been updated fairly recently to be inline with what our designer specified.

jonro2955:
@KevinMulhern I see. Thanks for looking into it. Will take this into consideration for next time.

PR #2818, opened 2022-04-04
Fix: Change SVG advert icons from Black to White in Dark Mode

ArudoPE:
I propose this as a fix to the Thinkful icons, because right now, they're black in dark mode, making them difficult to see. So this is a quick fix making them white in dark mode.

  - `Fix` - bug fixes

<hr>

**1. Because:**
When you're using Dark Mode on the Odin Project, the Thinkful icons are black, making it difficult to see, right now they look like this in Dark mode:
**Before**
before_theodin_icons

**2. This PR:**
I propose to change the color of the icons from black to white as a quick fix to this problem, and they'll look like this after the fix:
**After**
after_theodin_icons

### DIFF
--- a/app/assets/stylesheets/dark-mode.css
+++ b/app/assets/stylesheets/dark-mode.css
@@ -135,3 +135,7 @@ progress-circle__inner {
 .progress-circle__content {
   color: var(--dark-bg-accent) !important;
 }
+
+.advert__icon {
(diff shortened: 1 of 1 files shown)


KevinMulhern (approved):
Looks great, thanks again @ArudoPE

PR #2890, opened 2022-05-07
Fix: Correct typo in lesson seed that broke github import

ChargrilledChook:
Teensy typo was causing curriculum import task to fail for Battleship content.

cc @01zulfi


01zulfi (approved):
Nice catch, thanks @ChargrilledChook :rocket:

PR #2894, opened 2022-05-08
Chore: Update and improve bin scripts

ChargrilledChook:
-   [x] The title of this PR follows the `keyword: brief description of change` format

<hr>

**1. Because:**
It provides several optional but convenient shortcuts that streamline several common workflows

**2. This PR:**
* Regenerate default rails binstubs
* Tweak setup and update scripts to use yarn and run curriculum import task
* Add foreman convenience stub (copied verbatim from Rails 7 default)
* Add script to run all linting and test tasks

If people like this I think it could be added to the wiki as a way to streamline the process for people wanting to run the app locally. My guess is getting postgres started will be the tricky part for most people, but this streamlines everything else.

Everything except the test / linting task is a tweak of the default Rails scripts. 

There's one (very minor) issue with the setup script - in theory it should still work without issues on an existing database, but it won't rerun seeds. Just a limitation of how our seeds are setup and how `db:prepare` works. Still valuable and works for the intended purpose.

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+if ! command -v foreman &> /dev/null
+then
+  echo 'Installing foreman...'
+  gem install foreman
+fi
+
+foreman start -f Procfile.dev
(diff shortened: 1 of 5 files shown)


KevinMulhern (approved):
Thanks for taking the time to put this together @ChargrilledChook 💪 

These all work brilliantly, love the idea of streamlining the set up guide on the wiki with these.

PR #2990, opened 2022-06-18
Chore: Add null constraint to course badge_uri field

ChargrilledChook:
Because:
* Originally done in PR #2889, but need to stagger to avoid
a chicken/egg problem when changing the null constraint

This commit:
* Add a null constraint now all courses in prod have a non null value

Follow up from #2889 / #2883


ChargrilledChook:
I think we're safe to add this without an extra step of setting defaults etc now the previous change has gone through? Correct me if I'm wrong :sweat_smile:

KevinMulhern (approved):
Looks great! 🚀

PR #2991, opened 2022-06-18
Chore: Update Rspec config with example persistence

ChargrilledChook:
Because:
* It allows us to use the handy `--only-failures` feature of Rspec

Credit to the book 'Effective Testing with Rspec 3' for the idea

Usage: After updating, run specs at least once. You can now pass `--only-failures` or `--next-failure` to only run failed specs from the last run. `--next-failure` will bail after hitting one failure whereas `--only-failures` will run all previously failed specs

```
# Example usage
bundle exec rspec --only-failures
bin/rails spec:jobs --next-failure
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,6 @@ pickle-email-*.html
 /yarn-error.log
 yarn-debug.log*
 .yarn-integrity
+
+# Spec example persistence / failure tracking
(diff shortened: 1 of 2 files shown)


KevinMulhern (approved):
Nice! this is going to come in handy @ChargrilledChook ++

PR #2994, opened 2022-06-19
Feature: Store Announcement Creator

### LINKED ISSUE #2993: Feature: Store Announcement Creator
### Description
We should store who creates announcements for auditing purposes.

We should add a `created_by` column to announcements that will automatically be populated when creating an announcement.

### Acceptance Criteria
1. Go to admin
2. Create an announcement
3. It will redirect you to the announcements index page
4. The "Created By" column should contain your user name.

ChargrilledChook:
**1. Because:**

Closes #2993 

**2. This PR:**
* `Announcment` now `belongs_to` `user`
* `Announcement` admin form has hidden field to track `user_id`
* Add index view to active admin with `Created by` column + sorting

First time working with Active Admin, fairly annoying :laughing: 
If this approach is ok, I can add a bit of extra polish - adding a user field to the search filter and fixing up the naming conventions (`Created By` wants to default to `User` due to how associations are set up, and manually configuring is somewhat tedious.

**Todo**
- [x] Basic functionality working / acceptance criteria complete
- [x] Figure out strategy for existing announcements (optional? A default user?)
- [x] Squash / fix commit format

**Bonus**
- [x] ~~Change foreign key to `created_by` ?~~
- [x] ~~Polish - add `created_by` to filter, possibly show page~~

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,6 +16,7 @@ class User < ApplicationRecord
   has_many :user_providers, dependent: :destroy
   has_many :flags, foreign_key: :flagger_id, dependent: :destroy, inverse_of: :flagger
   has_many :notifications, as: :recipient, dependent: :destroy
+  has_many :announcements, dependent: nil
   belongs_to :path, optional: true
 
   def progress_for(course)
--- a/app/admin/announcements.rb
+++ b/app/admin/announcements.rb
@@ -1,10 +1,24 @@
 ActiveAdmin.register Announcement do
-  permit_params :message, :expires_at
+  permit_params :message, :expires_at, :user_id
+
+  index do
+    selectable_column
+    id_column
+
+    column :created_at
+    column :updated_at
+    column :message
+    column :expires_at
+    column 'Created By', :user_id, sortable: :user_id do |announcement|
+      announcement.user&.username || '🤷'
+    end
+  end
 
   form do |f|
     f.inputs do
       f.input :message
       f.input :expires_at, as: :datepicker
+      f.input :user_id, input_html: { value: f.current_user.id }, as: :hidden
     end
     f.actions
   end
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -4,4 +4,6 @@ class Announcement < ApplicationRecord
 
   scope :unexpired_messages, -> { where(expires_at: Time.zone.now..Float::INFINITY).order(created_at: :desc) }
   scope :showable_messages, ->(disabled_ids) { unexpired_messages.where.not(id: disabled_ids) }
+
+  belongs_to :user, optional: true
 end
--- a/db/migrate/20220621121447_add_created_by_to_announcements.rb
+++ b/db/migrate/20220621121447_add_created_by_to_announcements.rb
@@ -1,0 +1,5 @@
+class AddCreatedByToAnnouncements < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :announcements, :user, foreign_key: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_06_19_133130) do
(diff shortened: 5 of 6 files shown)


ChargrilledChook:
Demo: 
image

db/migrate/20220619021917_add_created_by_user_to_announcement.rb
```diff
@@ -0,0 +1,5 @@
+class AddCreatedByUserToAnnouncement < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :announcements, :user, null: false, foreign_key: true
```
KevinMulhern, line 3: We'll have issues with the null false and the existing announcements. But we can do the same as we did the last time by removing the null false, merging, updating the existing announcements to be created by one of our accounts and then re-add the null false constraint.

KevinMulhern:
> If this approach is ok, I can add a bit of extra polish - adding a user field to the search filter and fixing up the naming conventions (Created By wants to default to User due to how associations are set up, and manually configuring is somewhat tedious.

The approach is solid 👌 we're planning on ditching active admin in favour of rolling our own. We're better to make the switch before we build out a lot with active admin and rely on it too much. With that in mind, I don't think it's worth the effort of doing a lot of polish. What we have here is perfect mate 😄

app/models/user.rb
```diff
@@ -16,6 +16,7 @@ class User < ApplicationRecord
   has_many :user_providers, dependent: :destroy
   has_many :flags, foreign_key: :flagger_id, dependent: :destroy, inverse_of: :flagger
   has_many :notifications, as: :recipient, dependent: :destroy
+  has_many :announcements, dependent: nil
```
KevinMulhern, line 19: This will raise an exception when trying to delete an admin user with announcements because of the foreign key. But I think that's perfectly fine for the moment; a feature that solves the problem of being able to delete admin users, not a bug 😄  Just something we will need to take into consideration with the new admin.

ChargrilledChook:
Glad to hear! First time working in active admin, can't say I love it. Doesn't seem like a great use case for a DSL to me but there's probably a bit of historical context :shrug:

app/models/user.rb
ChargrilledChook: > a feature that solves the problem of being able to delete admin users, not a bug smile

Love this line of thinking :laughing: 

I figured nullify defeated the purpose of tracking announcement owners, and destroying announcments when deleting an admin even more so. Although I figure hard deleting an admin account is probably not something that happens that often?

db/migrate/20220619021917_add_created_by_user_to_announcement.rb
ChargrilledChook: Any thoughts on what to do with the historical ones? Is there a generic super user admin account they can be tied to?

KevinMulhern: There isn't any generic super users unfortunately, we could create one or associate them with one of the existing admin accounts for now if that's easier? I can sort this out through the Heroku console if you want?

app/models/user.rb
KevinMulhern: Yeah, you're totally correct, it doesn't happen that often. A separate admin_users table in the future so we can soft delete by deactivating an admin instead of deleting will be the ultimate solution for this.

db/migrate/20220619021917_add_created_by_user_to_announcement.rb
ChargrilledChook: Open to ideas, but I think probably seems like the simplest solution outside of making the relationship optional and not forcing the null constraint. 

Probably better to tie them to a generic account just for the sake of accuracy, but I guess it depends on how many actual records there are, is it worth the effort of tying existing ones to the right person?

ChargrilledChook:
@KevinMulhern I've removed the null constraint and made the FK optional for now but happy to revisit, figured I'd get something working out first.

One minor edge case I found was that editing an announcement will change the user / created by value to the editor. Probably not a serious one though, especially if ActiveAdmin is going the way of the dodo

KevinMulhern (approved):
Looks great, thanks @ChargrilledChook :shipit:

PR #3010, opened 2022-06-25
Chore: Remove controllers specs and gem (#3008)

### LINKED ISSUE #3008: Chore: Remove remaining rails controller tests
### Description
We have a very comprehensive system test suite which covers most if not all of the controller behaviour. The Rails team considers controller tests as depreciated so we should remove any we have remaining and the associated rails-controller-tests gem.

ChargrilledChook:
**Because:**
* Controller tests are deprecated

**This commit:**
* Remove rails-controller-testing gem
* Remove spec/controllers
* Fix broken request specs by removing render_template calls

**Additional Notes** 
Some of the request specs are probably of borderline utility now that `render_template` is gone but I'm not familiar enough with the feature specs to say for sure

Closes #3008


KevinMulhern (approved):
I think you're right, the request specs may need some rejigging to make sure they are testing the right thing. They should all be testing stuff that isn't easily tested with system specs.

Thanks for doing this @ChargrilledChook 💪

PR #3016, opened 2022-06-26
Chore: Remove old migrations (#3015)

### LINKED ISSUE #3015: Chore: Remove old Migrations
### Description
We recommend contributors to use `db:schema:load` when setting up the app because we have a few old migrations that are broken due to also being data migrations. 

We don't need to keep these old migrations when we have the option of using `db:schema:load`

### Acceptance Criteria
Remove any migrations more than 1 year old.

ChargrilledChook:
**Because**:
It's good to tidy

Closes #3015 

**Additional Information:**

The effort to satisfaction ratio of making this change was very high

delete


KevinMulhern (approved):
That lines added / deleted ratio is always satisfying to see 😍 

Nice work @ChargrilledChook 🚀

PR #3047, opened 2022-07-02
Chore: Switch to Pagy on the All Submissions Page (#3046)

### LINKED ISSUE #3046: Chore: Switch to Pagy on the All Submissions Page
### Description
We only have pagination on the [all submissions page](https://www.theodinproject.com/lessons/781/project_submissions) and are currently using kaminari for pagination, but [pagy](https://github.com/ddnexus/pagy) is better and more popular. We should adopt in now before we need pagination for something else such as the new admin in the near future.

ChargrilledChook:
Because:
* Pagy has better features and is more popular than Kaminari. Currently
the cost of changing is low.

This commit:
* Remove Kaminari pagination gem
* Replace with pagy

Closes #3046

**Note:**
Although I removed Kaminari from the Gemfile it's still a dependency of Active Admin.


app/controllers/lessons/project_submissions_controller.rb
```diff
@@ -7,10 +7,7 @@ class ProjectSubmissionsController < ApplicationController
     def index
       @user_submission = user_submission
 
-      @project_submissions = Kaminari.paginate_array(
-        public_project_submissions,
-        total_count: public_project_submissions.size
-      ).page(params[:page]).per(15)
+      @pagy, @project_submissions = pagy(public_project_submissions, items: 15)
```
ChargrilledChook, line 10: Defaults to 20 per page but can be tweaked in the config file

config/initializers/pagy.rb
```diff
@@ -0,0 +1,235 @@
+# frozen_string_literal: true
+
+# Pagy initializer file (5.10.1)
+# Customize only what you really need and notice that the core Pagy works also without any of the following lines.
+# Should you just cherry pick part of this file, please maintain the require-order of the extras
+
```
ChargrilledChook, line 6: Boilerplate from the installation guide. Haven't edited anything in here,

KevinMulhern (approved):
LGTM 💯 

Thanks @ChargrilledChook

PR #3066, opened 2022-07-06
Chore: Add null constraint to path badge uri (#3054)

ChargrilledChook:
Because:
* Paths should always have a badge. Follow up to #3053


KevinMulhern (approved):
LGTM 🚀

PR #3067, opened 2022-07-06
Fix: Temporary workaround for excessive JS errors in Sentry (#3048)

ChargrilledChook:
**Because:**
* The bug reduces our ability to see real errors

**This commit:**
* Implements one of the fixes suggested here https://github.com/getsentry/sentry-javascript/issues/5179

**Notes:**
* Unsure how to thoroughly test this aside from pushing it and seeing what happens, but open to ideas
* Should probably revisit once an official fix is found

### DIFF
--- a/app/javascript/packs/error_tracking.js
+++ b/app/javascript/packs/error_tracking.js
@@ -10,6 +10,12 @@ Sentry.init({
   allowUrls: [
     /https?:\/\/(www\.)?theodinproject\.com/,
   ],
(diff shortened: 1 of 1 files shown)


ChargrilledChook:
@KevinMulhern Cheers mate :tada: Given we can't really test this one directly, happy to push it to prod and keep an eye on it? If you want to merge it at a time that's better for you go ahead :ship:

KevinMulhern (approved):

PR #3150, opened 2022-07-30
Chore/update contribute page to tailwind

ChargrilledChook:
Closes theodinproject/top-meta#211

Because: 
We are updating all our styles to Tailwind

### Todo

- [x] Squash commits
- [x] CSS suggestions x3
- [x] Move erb hashes to translation file
- [x] Fix minor spacing issues in Hall of fame section
- [x] Fix Ryan's image being smaller than everyone else (? :thinking: )

### Extras (?)

- [x] Extract MediaCard Component -> issue raised, see #3379
- [x] Namespace translation keys -> see #3378


app/views/static_pages/contributing.html.erb
```diff
@@ -1,16 +1,34 @@
...
 </div>
 
-<%= render 'static_pages/contributing/benefits' %>
+<section class="bg-gray-100 odin-dark-bg-accent">
+  <div class="container">
+    <div class="grid grid-cols-12">
+      <div class="col-span-12 md:col-start-3 md:col-span-8">
+        <div class="page-container">
+          <h2 class="text-gold font-bold text-center pb-16 mb-0 text-2xl">Why you should get involved</h2>
+          <%= render(AboutPage::BeliefComponent.with_collection(t("contributing_page_beliefs"))) %>
```
ChargrilledChook, line 27: So I know this looks a little funky at the moment, but wanted to get some feedback before going all in on it. @leosoaivan 's component had almost the exact same layout as the one on this page, so I thought it would be a good spot for some re-use. 

If we're happy with this the next step would probably be to give it a more generic name, like `ImageCardComponent` or something similar. 

Only disadvantage of this is previously it was a 2x2 grid, now it's always stacked.

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -108,32 +108,25 @@
...
+    <div class="text-center underline">
+       <%= link_to 'View All Contributors', github_link('theodinproject/graphs/contributors'), class: 'text-gold mb-6', target: '_blank', rel: 'noreferrer' %>
+    </div>
   </div>
-
-  <%= render 'shared/bottom_cta',
-    button: chat_button,
-    heading: 'Have a question?',
-    sub_heading: 'Chat with our friendly Odin community in our Discord chatrooms!'
-  %>
```
ChargrilledChook, line 138: Deleted this because it overlaps with the updated footer:

image

This example is from the profile page, but same idea. Chat link only shows when logged in, though

app/views/static_pages/contributing.html.erb
KevinMulhern: Great spot with these components being the same 💪 

100% agree this is a great place for reuse. This looks like a media object to me:
https://tailwindui.com/components/application-ui/layout/media-objects
https://getbootstrap.com/docs/4.0/layout/media-object/

Perhaps that's the kind of naming we could use for the new component? -`MediaComponent`

> Only disadvantage of this is previously it was a 2x2 grid, now it's always stacked.
No worries with this, I think it's actually much easier to read when stacked in one column in that second screenshot.

app/views/static_pages/contributing/_hall_of_fame.erb
KevinMulhern: Makes complete sense to me, nice work!

app/views/static_pages/contributing.html.erb
KevinMulhern: In each of the icons, one of the path elements will have a fill attribute with a value of `#F3F3F3`
Changing those values to `none` will get rid of the white backgrounds.

Screenshot 2022-10-02 at 12 16 00

The clock looks a little bit odd, but we can always come back and tweak these later.

app/views/static_pages/contributing.html.erb
```diff
@@ -1,16 +1,34 @@
...
 
-  <%= render 'static_pages/contributing/contributing_options' %>
+      <%= render 'static_pages/contributing/contributing_options' %>
+    </div>
+  </div>
 </div>
 
-<%= render 'static_pages/contributing/benefits' %>
+<section class="bg-gray-100 odin-dark-bg-accent">
+  <div class="container">
```
KevinMulhern, line 22:
```suggestion
  <div class="page-container">
```

app/views/static_pages/contributing.html.erb
```diff
@@ -1,16 +1,34 @@
...
+    <div class="grid grid-cols-12">
+      <div class="col-span-12 md:col-start-3 md:col-span-8">
+        <div class="page-container">
```
KevinMulhern, line 25: We should be able to remove this div with the page container above being in place.

config/locales/en.yml
```diff
@@ -104,3 +104,20 @@ en:
     - image: 'img-connect.svg'
       subtitle: 'Connect'
       description: You're not alone. Learn and get help from our friendly community of beginner and experienced developers.
+  contributing_page_beliefs:
+    - short_text: Flexibility
```
KevinMulhern, line 108: Nit: What's your thoughts about changing all `short_text` instances between this and the about page beliefs to `title` instead? 

I know you got the `short_text` naming from the about page beliefs to keep the naming consistent and it's certainly not a blocker. I just noticed myself stoping to think about what this attribute represented.

KevinMulhern (changes requested):
Nice work @ChargrilledChook, this is looking great. Just a few minor suggestion.

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -108,32 +108,25 @@
   ]
 %>
 
-<div class="hall-of-fame gradient">
-  <h2 class="hall-of-fame__title text-2xl">Hall of Fame</h3>
-  <p class="hall-of-fame__text">
+<div class="pt-16 gradient">
+  <h2 class="text-gold text-center font-bold pb-7 text-2xl">Hall of Fame</h3>
+  <p class="text-center w-4/5 sm:w-3/5 m-auto mb-5">
```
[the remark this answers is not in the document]
ChargrilledChook: Ah, that is much nicer :heart: . I had a feeling it was a bit janky but was probably inspired by the previous implementation too much

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -108,32 +108,25 @@
...
     The Hall of Fame members are the hardworking devs who have built The Odin Project into the site that you see before you now.
   </p>
 
-
-  <div class="container hall-of-fame__members">
-    <div class="row">
+  <div class="container py-14 mb-6">
+    <div class="grid grid-cols-12 gap-6 mb-12">
       <% members.each do |member| %>
```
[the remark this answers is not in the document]
ChargrilledChook: Yeah good call, came close to doing this the first time round but there was slightly more overhead than the other one. Will do

config/locales/en.yml
ChargrilledChook: Good idea :bulb:, I think that reads much nicer

app/views/static_pages/contributing.html.erb
ChargrilledChook: Ah, nice one. I did attempt to change some fill values but obviously not that one :sweat_smile:

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -108,32 +108,25 @@
...
+<div class="pt-16 gradient">
+  <h2 class="text-gold text-center font-bold pb-7 text-2xl">Hall of Fame</h3>
+  <p class="text-center w-4/5 sm:w-3/5 m-auto mb-5">
     The Hall of Fame members are the hardworking devs who have built The Odin Project into the site that you see before you now.
   </p>
 
-
-  <div class="container hall-of-fame__members">
-    <div class="row">
+  <div class="container py-14 mb-6">
```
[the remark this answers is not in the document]
ChargrilledChook: Ah, assumed the Tailwind one was taking precedence - that explains  a few things

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -1,139 +1,31 @@
...
+<section class="pt-16 gradient">
+  <div class='page-container'>
+    <h3 class="text-gold text-center font-bold pb-7 text-2xl">Hall of Fame</h3>
+    <p class="text-center max-w-prose mx-auto mb-6">
+      The Hall of Fame members are the hardworking devs who have built
+      The Odin Project into the site that you see before you now.
+    </p>
 
-<div class="hall-of-fame gradient">
-  <h2 class="hall-of-fame__title text-2xl">Hall of Fame</h3>
```
ChargrilledChook, line 112: Didn't notice the first time through, but there was a sneaky <h2.5> in here :smile:

config/locales/en.yml
```diff
@@ -104,3 +104,102 @@ en:
...
+      path: curriculum
+    - title: Contribute to the Main Site
+      description: |
+        The site you are on right now is a Ruby on Rails application. We are
+        always working on new features that improve the platform and help students in
+        their learning journey. As with any Open Source project, there is also a constant
+        backlog of outstanding issues and bugs that need to be fixed. Anyone is welcome
+        to join in and start working on any of the issues we currently have posted in
+        the Github repository for the site.
+      path: theodinproject
```
ChargrilledChook, line 205: @KevinMulhern Now that the the translation file is seeing a bit of use, thoughts on refactoring to use proper translation key structure throughout? ie 

`en.static_pages.contributing.contributing_options.ways_to_contribute`

for this particular case / etc. 

A little bit of extra overhead but lays a good pattern for future use. Can do this in a separate ticket if we're keen

config/locales/en.yml
```diff
@@ -104,3 +104,102 @@ en:
...
+    - name: Leo
+      image: https://avatars2.githubusercontent.com/u/13614862?v=3&s=400
+      github_url: https://github.com/leosoaivan
+    - name: Chad Kreutzer
+      image: https://avatars1.githubusercontent.com/u/13308917?v=3&s=400
+      github_url: https://github.com/ChadKreutzer
+    - name: Roberto Pando
+      image: https://avatars1.githubusercontent.com/u/20132181?v=3&s=400
+      github_url: https://github.com/RobPando
+    - name: Ryan
```
KevinMulhern, line 185: I'm still seeing a smol Ryan avatar, but swapping him and Rob so Ryan is second last in the list seems to sort it out

app/views/static_pages/contributing/_hall_of_fame.erb
```diff
@@ -1,139 +1,30 @@
...
+    </p>
 
-
-  <div class="container hall-of-fame__members">
-    <div class="row">
+    <div class="grid grid-cols-12 gap-6">
+      <% members = t('hall_of_fame') %>
       <% members.each do |member| %>
-        <div class="col-lg-2 col-md-3 col-sm-4 col-xs-6 hall-of-fame__member">
+        <div class="col-span-4 sm:col-span-3 md:col-span-2">
```
KevinMulhern, line 13: The smallest and largest breakpoints don't appear to be kicking in, I tried a few different things but can't say for sure why they aren't working. It might be the gaps. But these breakpoints/col-spans work fine: 

```suggestion
        <div class="col-span-6 sm:col-span-3 lg:col-span-2">
```

config/locales/en.yml
KevinMulhern: 💯 good idea, the file is getting a bit unwieldy aha

KevinMulhern (changes requested):

app/views/static_pages/contributing/_hall_of_fame.erb
ChargrilledChook: Ah bugger, these were both working before I pushed them, but now I'm checking them locally they're both broken there too. The classes don't seem to be being picked up, even when I roll back to a commit where I know they were working.

Strangely, at one point I couldn't get a margin utility to work (`mb-5`) but changing it to `mb-6` and it kicked in. 

Same thing seems to be happening here, classes just aren't being picked up at all. I'll try and have a play around with it and see if I can find the cause

KevinMulhern: Ahh, I know what it is. The partial filename doesn't have a `.html` extension

ChargrilledChook: Ah, classic, always something dumb :lol: good eyes mate :pray:

I was getting there, I'd just managed to get the bundle to pick up the classes by pasting the component in the main template

ChargrilledChook: I wonder how long that had been hiding in there!

app/views/static_pages/contributing/_contributing_options.html.erb
```diff
@@ -1,35 +1,26 @@
...
-      path: "curriculum",
-    },
-    {
-      title: "Contribute to the Main Site",
-      description: "The site you are on right now is a Ruby on Rails application. We are always working on new features that improve the platform and help students in their learning journey. As with any Open Source project, there is also a constant backlog of outstanding issues and bugs that need to be fixed. Anyone is welcome to join in and start working on any of the issues we currently have posted in the Github repository for the site.",
-      path: "theodinproject",
-    }
-  ]
-%>
+<div>
```
ChargrilledChook, line 1: Previously the spacing between sections was pretty massive, originally I recreated it but I've gone through and given it a clean to try and rely more on the `page-containers` as you've suggested to get more consistent spacing. I think it looks better this way, but let me know if you think this has gone too far the other way

image

app/views/static_pages/contributing/_hall_of_fame.erb
KevinMulhern: > Maybe related to the Tailwind 3 upgrade?

Great point, you're absolutely right, It is related. TW3 uses JIT to load in classes whereas TW2 loaded everything in during development and purged any unused classes during the production build.

config/locales/en.yml
ChargrilledChook: Working now post the `.erb` fix :tada:

KevinMulhern (approved):
Looks great @ChargrilledChook, nice work mate! 💪

PR #3378, opened 2022-10-04
Chore: Restructure YAML keys in locale files

ChargrilledChook:
**Because:**
  The translation file was beginning to become unwieldy and could benefit from some organisation

**This commit:**
  * Namespaces keys in translation file in the way expected by I18n
  * Converts existing translations to use lazy look-up

**Notes:**
  Branched off of and pointed at #3150

### DIFF
--- a/app/views/static_pages/about.html.erb
+++ b/app/views/static_pages/about.html.erb
@@ -12,7 +12,7 @@
       </div>
 
       <h2 class="bold text-center text-2xl mb-10">Our Beliefs</h2>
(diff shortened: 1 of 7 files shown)


ChargrilledChook:
Another approach would be to break up the files by folder, not sure if it quite worth justifying yet since most of the content is under static pages anyway, but worth a though

KevinMulhern (approved):
Great stuff, thanks for doing this @ChargrilledChook 🥇

PR #3382, opened 2022-10-05
Fix: Incorrect translation key for about page beliefs

ChargrilledChook:
**Because:**

The key is wrong so the the title is not rendering.

**Notes**

Was originally correct, but a regression snuck in after a rebase. 
Find at `/about`. Should display like so:

image


KevinMulhern (approved):
:shipit:

PR #3406, opened 2022-10-10
Fix: White background in dark mode on homepage

ChargrilledChook:
**Because:**
* It's very bright and it burns my eyes

**This commit:**
* Replaces colour class with appropriate dark mode utility

**Notes:**
* Won't matter soon with proper native darkmode on the way, but an easy fix in the meantime

Before: 

image

After: 

image


01zulfi (approved):
:fire:s my :eye:s too

PR #3474, opened 2022-10-30
Chore: Remove New Relic config

ChargrilledChook:
**Because:**
* It is not being used


KevinMulhern:
Hey @ChargrilledChook, we're still using NewRelic.

It can only be through the app Dashboard on Heroku: 
Screenshot 2022-10-30 at 11 53 36

ChargrilledChook:
@KevinMulhern ah my bad, I thought this was handled by Sentry / the config looked ancient 🙈

KevinMulhern:
No worries mate, it's not that well known within the team. It's only used every so often for performance monitoring, but when it's needed, it's invaluable. Plus, we get a very generous free plan with Heroku.

PR #3591, opened 2023-01-15
Chore: Remove rails/webpacker package

ChargrilledChook:
Because:
* This package is EOL and we have migrated to Shakapacker


ChargrilledChook:
Removing it has broken some babel dependencies, converting to draft while I figure it out

KevinMulhern (approved):
Great catch @ChargrilledChook 💪

PR #3594, opened 2023-01-15
Chore: Add debug gem

ChargrilledChook:
Because:
* Currently there is no debugger for the app. The updated debug gem is now part of the standard library in our Ruby version, as well as being the default for Rails in v7.

These make it a sensible and safe choice to add as our default debugging gem.


ChargrilledChook:
I've discovered there's an incompatibility with this and the Rubymine debugger. Adding `require: false` seems to fix the issue, though

PR #3801, opened 2023-05-20
Chore: Add Capybara and Factorybot to Rubocop config

ChargrilledChook:
Because:
* They're already loaded as part of Rubocop-rspec. Silences recommendation to add when linting


KevinMulhern (approved):
🥳

PR #3807, opened 2023-05-21
Upgrade: Ruby 3.1 to 3.2

ChargrilledChook:
This commit:
* Bump Ruby minor version
* Update Gemfile.lock to use latest bundler version


ChargrilledChook:
This was intended to be a spike to see how much effort would be involved, but there doesn't appear to have been a single regression :tada:

ChargrilledChook:
One potential issue is [the removal of 3rd party sources from the Ruby install](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/#:~:text=Stdlib%20compatibility%20issues-,No%20longer%20bundle%203rd%20party%20sources,-We%20no%20longer), but given both CI and the review app worked this seems unlikely to be an issue for us

KevinMulhern (approved):
Looks great to me @ChargrilledChook 💪

PR #3821, opened 2023-05-27
WIP Spike for importing curriculum content from GitHub dynamically

ChargrilledChook:
- Add a gem to easily parse frontmatter from md files
- Attempt to create lessons via frontmatter
- Currently pointed at a specific dir on a specific branch of curriculum
- See https://github.com/TheOdinProject/curriculum/tree/spike/frontmatter-in-lessons/ruby/test_folder_please_ignore


ChargrilledChook:
Approach has some potential but would need a few improvements for feature parity - biggest problem that needs to be solved is how to delete lessons. 

Also worth considering broader approach to content management before implementing something like this

PR #3845, opened 2023-06-12
Feat: Enable Turbo Drive globally on site (#3806)

### LINKED ISSUE #3806: Chore: Complete Hotwire / Turbo Upgrade
## Context

We now patially use Turbo, but it is disabled by default for almost all pages. In order to make use of more features and modernise the app with current Rails 7 defaults it would be good to enable it globally. At the same time, we should clean up older Rails JavaScript on the app (ie, UJS) if and where possible.

## Tasks

- [ ] Toggle Turbo to be on on all pages by default
- [ ] Manually disable on any forms or links where appropriate
- [ ] Remove UJS from the app, and replace any behaviour as required
- [ ] Style the Turbo loading bar to use the site's colour theme (optional but nice to have)

## Acceptance Criteria

- [ ] Turbo is on globally
- [ ] UJS is removed from the app
- [ ] Rails tests continue to pass
- [ ] JavaScript tests continue to pass
- [ ] There are no regressions in site functionality

ChargrilledChook:
**Because:**
* It makes the site blazingly™️ fast
* Mostly closes #3806 

**This commit:**
* Remove global Turbo disable
* Add compatibility hacks for ReactRailsUJS to play nicely with Turbo

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -21,7 +21,7 @@ import 'hint.css/hint.min.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './src/js/axiosWithCsrf';
 import 'controllers';
-import { Turbo } from '@hotwired/turbo-rails';
+import '@hotwired/turbo-rails';
 
 Rails.start();
 
@@ -29,5 +29,7 @@ const componentRequireContext = require.context('./components', true);
 const ReactRailsUJS = require('react_ujs');
 
 ReactRailsUJS.useContext(componentRequireContext);
-
-Turbo.session.drive = false;
+ReactRailsUJS.handleEvent('turbo:load', ReactRailsUJS.handleMount);
+ReactRailsUJS.handleEvent('turbo:before-render', ReactRailsUJS.handleUnmount);
+ReactRailsUJS.handleEvent('turbo:frame-load', ReactRailsUJS.handleMount);
+ReactRailsUJS.handleEvent('turbo:frame-render', ReactRailsUJS.handleUnmount);


app/javascript/application.js
```diff
@@ -21,13 +21,15 @@ import 'hint.css/hint.min.css';
...
 const componentRequireContext = require.context('./components', true);
 const ReactRailsUJS = require('react_ujs');
 
 ReactRailsUJS.useContext(componentRequireContext);
-
-Turbo.session.drive = false;
+ReactRailsUJS.handleEvent('turbo:load', ReactRailsUJS.handleMount);
+ReactRailsUJS.handleEvent('turbo:before-render', ReactRailsUJS.handleUnmount);
+ReactRailsUJS.handleEvent('turbo:frame-load', ReactRailsUJS.handleMount);
+ReactRailsUJS.handleEvent('turbo:frame-render', ReactRailsUJS.handleUnmount);
```
ChargrilledChook, line 35: I considered adding a feature flag for this, but based on my manual testing / test suite it doesn't seem to be necessary. Happy to add one if we want to roll out in a safer way

CouchofTomato:
@KevinMulhern 

Seems ok to me and tests are passing. Any issues you know about?

KevinMulhern (approved):
Looks great @ChargrilledChook, blazing fast 🐎 

Just a couple of minor things:

 Do we need "data-turbo-track": "reload" on the stylesheet link tags?

Theres a little bit of strange caching behaviour when submitting a project 
1. Submit a project
1. Go to the next lesson
1. Go back in the browser
1. 💥 Project isn't there

A page refresh will sort it out, but some users often forget to try that.

We used to get loads of reports about the complete button inexplicably changing back to the previous state - ended up being the same scenario as this, only with Turbolinks instead lol.

turbo-permanent on the react element might sort it out, but it's not a show stopper without it. Just something to keep in mind if we get any reports of disappearing submissions.

ChargrilledChook:
> Do we need "data-turbo-track": "reload" on the stylesheet link tags?

My understanding is this means stylesheets will [only be reloaded if content changes](https://turbo.hotwired.dev/handbook/drive#reloading-when-assets-change) and needs to be re-fetched. It's also the default in new Rails 7 apps.

Given that seems like it would be better to leave it in, but let me know if I've missed something. Happy to take it out if needed

> Theres a little bit of strange caching behaviour when submitting a project

Good catch, I tried to beat up the React components as much as possible but didn't think to check that. As you say I think it's a relatively minor issue, but I'll follow up with a fix for this

PR #3906, opened 2023-07-02
Chore: Switch to route based error handling

### LINKED ISSUE #3876: Chore: (re)consider how TOP handles error pages
-   [x] The title of this issue follows the `brief description of request` format, e.g. `Add dark mode to website`

-   [x] I would like to be assigned this issue to work on it

<hr>

**1. Description of the Feature Request:**

Ruby on Rails offers three ways of handling error pages:
1. Static pages in the `/public` folder
2. A custom defined controller
3. Route based matching and mappping via `routes.rb`

Currently TOP uses a mix between 1 and 2; `422.html` and `500.html` are static pages in `/public`, while `404.html` gets handled with a custom controller.

Both of these approaches however offer disadvantages:
* Pages in the `/public` folder are outside the app scope, so they can't be `.erb` files or use tools such as Tailwind, necessitating hardcoding links and CSS.
* While the custom controller as such is not bad, its essentially just a complex way of achieving the same effect as directly using `routes.rb`

Moving to route based handling would simplify both of these, but also offers risks: if something else that that code relies on also contains a error, Rails will be unable to render custom pages, and will instead emit plaintext error pages. Per @KevinMulhern  however, this only affects exceptions raised from within app itself, as Heroku already handles it otherwise. This risk should therefore be minimizable by reducing the amount of shared code and having full test coverage  

**2. Acceptance Criteria:**
- [ ] A single way to handle error page is decided upon
- [ ] All error pages use this method

Asartea:
## Because
- Currently we use two different ways of handling errors: a complex custom handler that really is just matching it to a specific controller, and static error pages. We can handle errors more cleanly and uniformly by switching all error pages to instead use Rails route based handling, which lets us map `/STATUS_CODE` requests to specific controller actions

## This PR
- Switches our error handling method to routes
- adds routes in `routes.rb` for `422` and `500`
- adds a controller and rewrites `ErrorsController` to handle these: `ErrorsController` handles `500` and `422` and `NotFoundController` handles `404`
    - I chose this approach because it enables us to separate `500` and `422` from the application logic, while `NotFoundController` still inherits from `ApplicationController`, and therefore gets to keep the nice default TOP layout, as its much more likely to be encountered
- adds tests for checking that each of the error pages return the correct status code (and ergo work), and an additional test to check that `500` actually returns our custom page and not just the fallback page.
- drops the now unnecessary `/public/500.html`, `/public/422`, and `/lib/custom_public_exceptions` files
   

## Issue

Closes #3876 

## Pull Request Requirements

  - `Fix` - bug fixes


app/controllers/not_found_controller.rb
```diff
@@ -0,0 +1,5 @@
+class NotFoundController < ApplicationController
+  def not_found
```
KevinMulhern, line 2: Nit: We can move this to the errors controller and use the application layout just for this action:

```ruby
class ErrorsController < ActionController::Base
  ...
  def not_found
     render status: :not_found, layout: 'application'
  end
```

Main benefit would be having all error actions in the one controller so they easier to find and extend.

app/views/errors/internal_server_error.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= title('Internal Server Error (500)') %>
+
+<p class="text-base text-gold-600">500</p>
+<h1 class="mt-4 text-5xl text-slate-900 font-bold">Server error</h1>
+<p class="leading-7 mt-6 text-slate-600">We're sorry, but something went wrong.</p>
+<div class="mt-10 flex items-center justify-center gap-x-6">
+    <a href="/home" class="rounded-md bg-teal-700 px-3.5 py-2.5 text-sm text-white shadow-sm hover:bg-teal-800 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-teal-600">Go back home</a>
```
KevinMulhern, line 7: Nit: Indentation seems off here, 2 spaces is our preferred indentation.

KevinMulhern, line 7: Nit: We can make a couple of QOL improvements to this now that all of Rails/ Tailwind is available to us: 
* We can use a dynamic link_to here, it will take users to their dashboard if logged in / the home page if not. 
* We can also use our button styling here too.
* This will need updated in the 422 page too

```suggestion
    <%= link_to "Go back home", root_path, class: "button button--primary" %>
```

config/application.rb
```diff
@@ -15,8 +15,7 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
-    require Rails.root.join('lib/custom_public_exceptions')
```
KevinMulhern, line 18: 🎉 great to see this going

spec/system/errors/errors_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'rails_helper'
+
+describe 'GET 500' do
+  it 'displays the page' do
```
KevinMulhern, line 4: Nit: This is a really good thing to test. But doing it through the request specs is the preferred place. One for each of the pages to ensure they render the correct template in the `requests/error_spec.rb`:

```ruby
it 'renders the correct template' do
    get '/500'
    expect(response.body).to include("We're sorry, but something went wrong.")
end
```

System tests in Rails are pretty slow and because of that are mostly reserved for testing happy path workflows through the app. Making sure a user can navigate to a course and complete a lesson kind of stuff.

KevinMulhern (changes requested):
I just have a few minor things for feedback, but this is looking mighty solid. Nice work @Asartea 💪

app/controllers/not_found_controller.rb
Asartea: > Nit: We can move this to the errors controller and use the application layout just for this action:

We can't, or at least not easily; the application layout intrinsically depends on the methods of ApplicationController, and just setting it here just results in a bunch of internal server errors and unknown/missing methods

KevinMulhern: Ahh ok, sorry, when testing locally I forgot to switch the route over lol

In that case, I think we're still best to aim for consistency with these by losing the navbar and footer for the 404 page and just putting it all through the errors layout.

We could maybe make the link back to home clearer but other than that it doesn't look bad:
Screenshot 2023-07-03 at 06 15 46

Another option would be giving the same kind of styling as 500/422 pages so they are completely consistent. But using the current styling isn't a deal breaker.

Asartea:
Should all be done: For the 404 page I opted to keep the current styling(mostly because I really like it), but did change the link styling for both to a more dark slate, which IMO looks cleaner than the almost but not quite black gold we had before, and repointed the head back home link to the root path (so /dashboard for logged in) instead of just to `/home`

KevinMulhern (approved):
Thanks @Asartea, this is looking great. I'm about to head off for the evening, but will merge and deploy first thing tomorrow.

Nice work with these pages 💪

PR #3915, opened 2023-07-03
Feature: Increase light mode background color

Asartea:
## Because
- Currently the various note stylings have a background color which is easily distinguishable on dark mode, but basically invisible on light mode, as the contrast between the note background and the page background is basically zero (1.04:1).

## This PR
- Increase the color to `bg-gray-200`, which produces a much clearer background that better distinguishes the note

## Pull Request Requirements

  - `Fix` - bug fixes

### DIFF
--- a/app/javascript/stylesheets/lesson-content.css
+++ b/app/javascript/stylesheets/lesson-content.css
@@ -24,7 +24,7 @@
   }
 
   .lesson-note {
-    @apply bg-gray-50 p-4 border-l-4 border-gold dark:bg-gray-700/40 dark:border-gold-700 rounded;
+    @apply bg-gray-200 p-4 border-l-4 border-gold dark:bg-gray-700/40 dark:border-gold-700 rounded;
   }
 
   .lesson-note > h4::before {


KevinMulhern (commented):
Thanks for taking the time to do this.

I think `bg-gray-200`  may be too distracting and break readability of content slightly:
Screenshot 2023-07-03 at 22 54 25


`bg-gray-100` looks better to me, But even that might be too much.
Screenshot 2023-07-03 at 22 53 49

IIRC, the choice of using a subtle background was very deliberate. With notes being located in the middle of larger blocks of text, a subtle background helps blend them into the overall content and prevent learners being taken out of their flow when reading. 

It basically needs to stand out enough to be noticeable - but not enough to make reading the content around it more of a strain. The side border with a very light background _should_ be enough to do that. The test is to read the text above/below and be aware of where your eyes are drawn too. 

But... I appreciate this is pretty subjective stuff and would like to get a few more opinions on this.

Asartea:
> IIRC, the choice of using a subtle background was very deliberate. With notes being located in the middle of larger blocks of text, a subtle background helps blend them into the overall content and prevent learners being taken out of their flow when reading.

> It basically needs to stand out enough to be noticeable - but not enough to make reading the content around it more of a strain. The side border with a very light background should be enough to do that. The test is to read the text above/below and be aware of where your eyes are drawn too.

I can see that, but currently I basically can't see the background at all, which is especially given the clearly visible dark mode one. Like for one of the attached images I have removed the background color entirely, and unless I actually strain my eyes I can't tell which one. 
image
image

KevinMulhern:
I do see where you're coming from, and you're not wrong - when the notes are in isolation like in those screenshots, the difference is almost unnoticeable. 

But notes are always used within larger blocks of content, and in that context the difference is noticeable. 
Screenshot 2023-07-04 at 09 47 45

As opposed to no background:
Screenshot 2023-07-04 at 09 48 21
 
Tbh, on light theme, the note background isn't that important. The border is what brings attention to notes, and the background is only really needed so the border doesn't look out of place.  

The dark theme is different because it doesn't work as well on that theme and [needs different design touches](https://twitter.com/steveschoger/status/1151160261170126850) to be effective. 

Ultimately, in the lesson content readability is the most important thing. For me, the current background provides the most comfortable reading experience, anything darker and my eye just wants to focus on it instead of what I'm actually supposed to be reading. But that _may_ just be me. @TheOdinProject/maintainers any thoughts?

ManonLef:
I too am in favour of keeping it the way it is. The screenshots Kevin shared really do show just enough contrast in my opinion to not make it stand out as a block so much as bg-gray-100 already does.

rlmoser99:
From my perspective, the background of the note being different than the assignment on light mode has stood out to me. I would lean towards having them be the same background color. On dark mode these two background colors are the same color, which I think looks cleaner.

At the same time, I totally understand that we need to treat light mode and dark mode differently. I wonder if this would be a good question to ask the designer of the website if they have a strong opinion either way.

ManonLef:
Screenshot 2023-07-04 at 16 23 17

That would be bg-gray-100
Here's what they look like if they have the same background

ManonLef:
Screenshot 2023-07-04 at 16 16 15

And how it is currently

ManonLef:
Screenshot 2023-07-04 at 16 28 16


On dark mode they're not the same I just noticed. 
assignment = bg-gray-800 
note = bg-gray-700/40

rlmoser99:
Good call out. They look a lot more similar on dark mode, so I just assumed that they were the same. I should have verified it. I think they should be the same, but this may be a whole different conversation.

ManonLef:
I think if it were to have no background color it could look good if the left border wasn't shaped with rounding. The rounding makes it seem as if it should wrap around something.

KevinMulhern:
gray-100 doesn't look bad at all with just the border. I think it's the icon thats making the other version weird with the content around it. A darker gray background contrasts it more and it just becomes a blob of color in the corner of your eye lol

Couple of ideas:

Smaller icon with a darker note accent color:
Screenshot 2023-07-04 at 17 27 29

Smaller, gray scale icon:
Screenshot 2023-07-04 at 17 28 01

Asartea:
At first glance I don't like the smaller corner; I feel like it makes it too cramped and I need to focus to easily discern it? Also looks weird next to the heading

ManonLef:
Screenshot 2023-07-04 at 19 02 35
Screenshot 2023-07-04 at 18 58 34
Screenshot 2023-07-04 at 18 57 47

I played a bit with rounded-sm and no rounding. In case you're curious how it'd look without a background and just a border left

KevinMulhern:
Thanks for mocking those up @ManonLef 💪 that doesn't look half bad. It's _maybe_ a little close to [our blockquote styling](https://www.theodinproject.com/lessons/foundations-problem-solving), but I like these versions best so far (besides our current design ofc lul 😈)

ChargrilledChook:
Not opposed to changes per se, but I'm not sure this is a problem that really needs solving either. Like @KevinMulhern mentioned I think readability is the first, second and third concern for this area. 

I haven't seen any feedback prior to this in (9 - 12 months?) of this style of note box either on the repo or in the Discord, so my intuition is that this is unlikely to negatively impacting many people's experience. Happy to hear evidence to the contrary, though. 

One related thing that would convince me this is worth improving would be accessibility concerns around contrast, but not an area of expertise for me. Could be worth running Lighthouse and seeing if it flags anything on the noteboxes

Asartea:
> One related thing that would convince me this is worth improving would be accessibility concerns around contrast, but not an area of expertise for me. Could be worth running Lighthouse and seeing if it flags anything on the noteboxes

I just checked and it shouldn't be a problem, although Lighthouse is flagging that (completely unrelated) our current light mode link color isn't meeting WCAG AA standards

KevinMulhern:
Thanks for reporting that back @Asartea. I think we should start a separate discussion about changing the link colors. There are some UX improvements beyond just improving accessibility we could gain from changing to a different color.

It would also be worth starting a discussion for these note design changes and closing this PR. We need to come to a consensus of what to do for it first and a discussion will be a better format for that.

PR #4240, opened 2023-10-26
Chore: curriculum task spec failures

JoshDevHub:
## Because

I often run the test suite excluding system specs (because I like to go fast), but I noticed that the test for the curriculum update task always fails when run this way. It also fails when run alone, which you should be able to verify by quickly running:
```
bundle exec rspec spec/tasks/curriculum_spec.rb
```
This seems to have emerged from #3893, which added precompiling assets when the test run includes system specs.

This test *does* work when the full suite is run, but it felt not great to me that it's dependent on the asset precompile happening.

## This PR

- Adds the usual rake task spec boilerplate to curriculum task spec. The absence of this is what causes it to fail when run without the asset precompile
- The above change gets the test passing when run without system specs (good) but now they fail when run with system specs (not so good). Kevin observes this here in the aforementioned [here in the aforementioned PR](https://github.com/TheOdinProject/theodinproject/pull/3893#discussion_r1244426518). Something about the task loading is interfering with this test. So instead of loading the tasks into the Rails app to invoke `assets:precompile`, I use `system` to run it as a shell command. All tests are green and assets still precompile for runs that include system specs. This should also still be fine on CI? Definitely let me know if there's some mistaken assumptions with that.
- Renames `spec/support/precomple_assets.rb` to `spec/support/precompile_assets.rb` (minor incidental typo correction)

## Issue

Closes #XXXXX

## Pull Request Requirements

    - `Fix` - bug fixes


KevinMulhern (approved):
Great stuff, thanks for digging into this @JoshDevHub

PR #4241, opened 2023-10-27
Feature: Add support for parsing backtick fenced codeblocks

JoshDevHub:
## Because

Kramdown's default supported syntax for delimiting codeblocks (fencing with three `~`) is pretty unconventional. This especially causes problems with VSCode's markdown editing, which has strong (negative) opinions of the tilde fenced codeblocks. This makes contributing more annoying because you have to pay attention to this quirk. It also makes maintenance more annoying because contributors may accidentally autoformat the tildes to backticks.

## This PR

- Add [kramdown-parser-gfm](https://github.com/kramdown/parser-gfm) gem to support "Github Flavored Markdown". One of the key features of this is support for backtick fenced codeblocks
- Edit the `MarkdownConverter` so that documents are parsed using github-flavored markdown

## Issue

Closes #XXXXX

## Additional Information

I've been unable to find any problems or new oddities caused by parsing with GFM. I reset my db and re-imported content locally using it and have been unable to find anything strange happening. But also I'm only one person and can't realistically browse everything. It's worth noting that the tilde syntax is still valid and will be correctly parsed, which is convenient for our usecase.

There are arguably some drawbacks to this. It's an extra dependency, and it's one that doesn't seem to get much attention. We could potentially look at rolling a custom solution for this, especially since the backtick codeblocks are the only thing I think we really care about.

This might be as simple as finding a way to modify [Kramdown's regex matching on codeblocks](https://github.com/gettalong/kramdown/blob/master/lib/kramdown/parser/kramdown/codeblock.rb#L33-L34).

Also if this is accepted, we probably want to spin up a bunch of tickets for converting the curriculum markdown documents over to using backtick fenced codeblocks. And also would need to update the style guide to tell contributors to use the backtick format.

## Pull Request Requirements

    - `Fix` - bug fixes


spec/services/markdown_converter_spec.rb
```diff
@@ -126,5 +126,22 @@
...
+      it 'converts the markdown with appropriate code format' do
+        markdown = <<~MARKDOWN
+          ```ruby
+          puts 'this should work'
+          ```
+        MARKDOWN
+
+        html_result = <<~HTML
+          <div class="language-ruby highlighter-rouge"><div class="highlight"><pre class="highlight"><code><span class="nb">puts</span> <span class="s1">'this should work'</span>
+          </code></pre></div></div>
```
JoshDevHub, line 140: This might look kind of strange if you're familiar with what actually gets rendered to the page at the end. But what gets rendered in the end is extensively decorated by the Prisma.js syntax highlighter library on the frontend. This is the correct format for what Kramdown parses out.

KevinMulhern (approved):
Nice work @JoshDevHub! it's working great for me. The legacy tildes still working with this makes things a whole lot easier 💪 

Would definitely be in favour of moving all code blocks towards backticks.

PR #4253, opened 2023-10-30
Chore: Don't restrict Rspec output length

Asartea:
## Because
- Rspec by default truncates the output of failed tests, but it doesn't do so in an intelligent way, instead just taking half the allowed characters from the front and the other half from the back, and replacing all characters in between with a single ellipsis. This means that with some test errors (such as the ones I encountered in #4242), it can be impossible to see which part of a string isn't matching.

## This PR
- Sets `max_formatted_output_length` to `nil`, which makes Rspec always print the full string

## Pull Request Requirements

    - `Fix` - bug fixes


KevinMulhern (approved):
TIL this is configurable - nice find @Asartea!

PR #4308, opened 2024-01-03
Backend: Basic plumbing for bookmark feature

ChargrilledChook:
A (very rough) spike for a lesson bookmarking feature to gain feedback on whether it's valuable and worth polishing.

Would need more thought around design, UI and UX. Ideally bookmarks would live on the existing dashboard. Would need to consider other technical limitations too (should users only be able to bookmark n lessons? If yes, how does adding and removing look at the limit?)

But before needing to answer all those questions, we can get test the waters and see if there's any appetite for this

### DIFF
--- a/app/components/bookmarks/button_component.html.erb
+++ b/app/components/bookmarks/button_component.html.erb
@@ -1,0 +1,10 @@
+<div class="flex items-center justify-center" id="bookmark-button">
+  <%= button_to create_or_destroy_path,
+                form_class: 'w-full h-full',
+                method: bookmarked? ? :delete : :post,
+                data: { test_id: 'bookmark-button' },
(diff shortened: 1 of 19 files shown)


app/components/bookmarks/button_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="flex items-center justify-center" id="bookmark-button">
+  <%= button_to create_or_destroy_path,
+                form_class: 'w-full h-full',
+                method: bookmarked? ? :delete : :post,
```
ChargrilledChook, line 4: Would have preferred to keep the logic in the VC class with the rest of it, but putting it in a method didn't work for some reason 🤷

Issue #4002, opened 2023-07-20
Bug: Broken home page banner - welcome email


mbiesiad:
Hi, referring to the fixed issue https://github.com/TheOdinProject/theodinproject/issues/3999 I noticed other issue there.

Complete the following REQUIRED checkboxes:

-   [x ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [ x] The title of this issue follows the `Bug: brief description of bug` format, e.g. `Bug: Lesson complete button does not update on click`

The following checkbox is OPTIONAL:

-   [ ] I would like to be assigned this issue to work on it

<hr>

**1. Description of the Bug:**
Home page banner looks broken - welcome message email.

**2. How To Reproduce:**
1. Go to https://www.theodinproject.com/sign_up and register in
2. Open email with welcome message from [contact@theodinproject.com](mailto:contact@theodinproject.com) titled "Getting started with The Odin Project"
3. Notice mentioned missing banner - see alt text

**3. Expected Behavior:**
Displayed home page banner in correct form.

**4. Desktop/Device:**
NA

**5. Additional Information:**
Screenshot - current result/view:
![image](https://github.com/TheOdinProject/theodinproject/assets/18367606/e79c0638-30c4-4ca8-b44a-6e6f18e17450)

Best wishes,

PR #4309, opened 2024-01-04
Fix broken home page banner

sparshalc:
This pr addressed #4002  . I have updated the relevant code and fixed the broken home page banner issue.


KevinMulhern (changes requested):
Hey @sparshalc, thanks for taking this one on 💪 .But,  we don't want to duplicating the logo svg code if we can help it. I think this is just a [configuration issue that should be easy enough to fix with a couple of lines of code](https://guides.rubyonrails.org/v5.2/action_mailer_basics.html#adding-images-in-action-mailer-views).

If we include the following In `config/environments/production.rb`:
```
config.action_mailer.asset_host = "https://www.theodinproject.com"
```

We should be able to render the image straight from our assets in the welcome.html.erb file:
```
<%= image_tag 'logo.svg', alt: 'home page banner' %>
```

PR #4406, opened 2024-02-23
Update convert_img to skip images with alt text of empty str

thatblindgeye:
## Because

Currently we wrap *every* image in an anchor element so that it can be opened in a new window, in order to potentially see a larger version of the image (if it's been scaled down to fit within the context of a lesson). There's going to be instances when we don't want to do that, such as when an image is purely decorative with an empty string alt text.

## This PR

- Updates the convert_img to skip images with `alt=""`
- Adds a test to ensure the above works as intended

## Issue

Closes #XXXXX

## Additional Information

@KevinMulhern lmk if there's a better way to grab the alt text of the img that's attempted to being converted in this function. 

There may be other issues we may want to consider with wrapping images in anchors in the first place:

- Is it clear that clicking the image opens the image in a new tab for the full sized version? Or could it be misconstrued as an image link, e.g. the TOP logo being an "image" that links to the home page.
- Would it be an issue that opening the image in a new tab results in the image having no `alt` attribute applied; this may be a bit out of our control, as even opening the image in a new tab via right clicking > "Open image in a new tab" results in this issue.

## Pull Request Requirements

-   [ ] I have thoroughly read and understand [The Odin Project Contributing Guide](https://github.com/TheOdinProject/theodinproject/blob/main/CONTRIBUTING.md)
-   [ ] The title of this PR follows the `keyword: brief description of change` format, using one of the following keywords:
    - `Fix` - bug fixes
-   [ ] The `Because` section summarizes the reason for this PR
-   [ ] The `This PR` section has a bullet point list describing the changes in this PR
-   [ ] I have verified all tests and linters pass after making these changes.

### DIFF
--- a/lib/kramdown/converter/odin_html.rb
+++ b/lib/kramdown/converter/odin_html.rb
@@ -8,7 +8,8 @@ module Kramdown
       EXTERNAL_LINK_ATTRIBUTES = { target: '_blank', rel: 'noopener noreferrer' }.freeze
 
       def convert_img(element, _indent)
-        return super if @stack.last.type == :a
+        
+        return super if @stack.last.type == :a or element.attr['alt'] == ""
 
         attributes = { href: element.attr['src'] }.merge(EXTERNAL_LINK_ATTRIBUTES)
         %(<a#{html_attributes(attributes)}>#{super}</a>)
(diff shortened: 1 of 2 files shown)


KevinMulhern (approved):
Change looks solid to me @thatblindgeye, nice work 💪

PR #4476, opened 2024-04-06
Upgrade: Ruby 3.3

ChargrilledChook:
Because:
- It's a stable change with minimal impact to the codebase, and we should keep up with the latest stable version of Ruby.

Changes:
- Remove ambiguous block forwarding inside spec
  - See https://rubyreferences.github.io/rubychanges/3.3.html#anonymous-parameters-forwarding-inside-blocks-are-disallowed
- Update CircleCI config to use Ruby 3.3


KevinMulhern (approved):
Looks great! thanks @ChargrilledChook

PR #4520, opened 2024-05-09
Feature: Adds realtime linting for markdown previewer.

### LINKED ISSUE #4340: Feature Request: Markdown preview tool linting
### Checks

- [X] The title of this issue follows the `Feature Request: brief description of feature request` format, e.g. `Feature Request: Add a dark mode to the website`

The markdown preview tool used when contributing to lessons (app/views/lessons/previews/) renders content without checking that it will meet the criteria applied by the linting tool in the new PR checks pipeline on the curriculum repo. This can lead to a situation where a user checks using the tool, everything looks fine, and then once the PR is open the bot will complain.

If this could be highlighted during the preview stage, it would prevent several commits, check what the bot says, commit again, etc

### Acceptance criteria

- [] Linting error highlighting in [markdown preview tool](https://www.theodinproject.com/lessons/preview)

### Additional Comments

I would offer to help on this if I had any idea how :laughing:

ZachBaird:
## Because
There is a [markdown preview page](https://www.theodinproject.com/lessons/preview). It allows contributors to see how the markdown will render on the site, but does not indicate any issues that may exist with the markdown. When pull requests are created in `curriculum`,  the linter detects these issues. The PR is prevented from merging, leading to additional commits satisfying linter conditions that the contributor was initially unaware of.

## This PR
This PR adds linting to the site. As users preview their markdown, a linter is run against it based on the `jsonc` configuration file from `curriculum`. Any potential issues are listed below the preview for the user to resolve.

Their markdown is loaded up into a temporary file that the linter is run against. Results are saved and the file is deleted immediately after. Results are then formatted for presentation on the page.

## Issue
Closes #4340 

## Pull Request Requirements
    - `Fix` - bug fixes

### DIFF
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -1,0 +1,84 @@
+// https://github.com/DavidAnson/markdownlint
+
+{
+  "config": {
+    "default": true,
+    // heading-style
+    // Enforces ATX style headings, e.g. ### A level 3 heading
+    "MD003": {
+      "style": "atx"
+    },
+    // ul-style
+    // Enforces dash syntax for unordered lists
+    "MD004": {
+      "style": "dash"
+    },
+    // line-length
+    // Avoids errors related to line lengths in the MD files
+    "MD013": {
+      "code_blocks": false,
+      "headings": false,
+      "tables": false,
+      // Arbitrary length - Revisit to possibly set a standard
+      "line_length": 9999
+    },
+    // no-trailing-punctuation
+    // Prevents most punctuation in headings, except for exclamation and question marks
+    "MD026": {
+      "punctuation": ".:,;"
+    },
+    // ol-prefix
+    // Enforces lazy numbering for ordered lists
+    "MD029": {
+      "style": "one"
+    },
+    // no-inline-html
+    // Only allows specified HTML to be inline
+    "MD033": {
+      "allowed_elements": [
+        "p",
+        "a",
+        "div",
+        "span",
+        "kbd",
+        "script",
+        "iframe",
+        "details",
+        "summary",
+        "pre"
+      ]
+    },
+    // first-line-heading/first-line-h1
+    // Enforces the first heading on a lesson page to be an H3, since the page template renders an H1 and H2 automatically
+    "MD041": {
+      "level": 3
+    },
+    // proper-names
+    // Enforces proper spelling of the names array, except for code blocks and HTML elements
+    "MD044": {
+      "code_blocks": false,
+      "html_elements": false,
+      "names": ["CSS", "HTML", "JavaScript"]
+    },
+    // code-block-style
+    // Prevents indented code blocks from being used
+    "MD046": {
+      "style": "fenced"
+    },
+    // code-fence-style
+    // Enforces backticks for codeblocks
+    "MD048": {
+      "style": "backtick"
+    },
+    // emphasis-style
+    // Enforces asterisk syntax instead of underscore syntax
+    "MD049": {
+      "style": "asterisk"
+    },
+    // strong-style
+    // Enforces asterisk syntax instead of underscore syntax
+    "MD050": {
+      "style": "asterisk"
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


ZachBaird:
This PR may need the server to have the buildpacks set up differently. It *should* work with something like the following:
```
heroku buildpacks:add --index 1 heroku/nodejs
heroku buildpacks:add heroku/ruby

heroku buildpacks # Check that node.js is executed before ruby
```

MaoShizhong:
How would we plan to account for custom rules? We currently have 7, with 1 in review (which requires a disabling of MD048) and 2 assigned in other issues (one of which is also a custom version of a built-in just to provide a fixer for our npm scripts in the curriculum repo).

ZachBaird:
@MaoShizhong I had a lot of uncertainty around this as well. My initial thinking was:

1. If we include just the base configuration file, it allows us to give a true preview on the site. It won't lint for everything `curriculum` does, but would catch the major stuff. It also allows us to pretty much ship once we sort out this pipeline error.
2. Including the custom rules right away would not be difficult, but just copying them over and maintaining them in 2 places isn't great. I'm also unsure where best to put the files (perhaps in `lib`). This was something I was going to consult @KevinMulhern on once he had a chance to review this.
3. We may need to be cautious about adding too much linting in one go. It's unclear how much of a performance impact this could have on the server if the previewer sees a spike in usage.

I think publishing it as its own npm package is a really great idea.

MaoShizhong:
@ZachBaird That's a very fair point about the performance impact we'd need to monitor.

I'm definitely not keen on the "duplicate the code and maintain in 2 places" thing, which was my first and biggest concern about feasability. That's fair though on the first point regarding not *needing* to lint everything. If people are fine with this then perhaps we could make it clear that the error list may not be exhaustive? I know 2 of the built-ins would cover 2 of the upcoming customs anyway, since those are customs purely because the built-ins don't have fixers but can have fixers, which would leave a potential 7 unchecked rules.

While I like the custom npm package idea, I am aware that'd mean we'd need to remind contributors that whenever they sync their fork's main with upstream, they should run `npm update` so they'd catch any package updates. While it seems relatively small to me, I'm aware this might cause some issues since custom markdownlint updates won't be brought along immediately by just syncing fork:main and upstream.

thatblindgeye:
Having the linting stuff housed in its own repo has been mentioned before, so how strongly we feel about implementing this sort of feature may force us to prioritize such an effort since - as mentioned - having the same content in 2 repos is not ideal and requires more upkeep than necessary. 

I'm probably on the fence of this integration. I think my main concern is whether the work required to get this working as ideally as possible makes sense in terms of how used the preview tool is. "Ideally as possible" would include transferring the linting content to its own repo/package as well as ensuring the preview tool can take into account custom rules/configs. The latter is important because if we based it solely on the base markdownlint config, it would most likely lead to confusion/contradiction when compared to our custom rules/config. Another thing to consider is that currently we have 2 configs: one for lessons and one for projects. So we'd need to be able to choose what config to run the preview against. Which if there's a worry about performance impact  or we need to limit what we lint, is it even worth implementing at that point?

Another thing to consider is what the scope of the preview tool should be. Should it take into account linting stuff, or should the scope be kept literally, i.e. "This tool will preview your markdown, nothing more". Linting the markdown feels like it goes a little beyond just "previewing", but we can also possibly tweak the intent of the tool. Rather than just "preview", maybe "validate" or "preview and lint". Could also just be semantics at that point, though.

That said, for the users who may not yet know how to use Node (or maybe just don't want to go through that hassle and just want to quickly copy paste something, like I've done at times), this would be a great improvement to the current preview tool.

mao-sz:
@KevinMulhern coming across this again, I echo @thatblindgeye's thoughts regarding the scope of the markdown preview tool. IMHO, it should stick to its name i.e. be a tool to preview the markdown on the website and only that.

Linting is out of scope of that and is something that maints can communicate with contributors in PRs if needed. Besides, if the preview tool did include linting, it *must always* be in sync with the curriculum repo linting, else it wouldn't actually be solving any problems. That introduces complexity - we'd need to extract custom linting to our own npm package, but then that makes it harder for people in the curriculum repo to keep in sync with that package. They'd have to manually `npm update` or we'd have to include a forced update in the linting scripts which is iffy. So in my eyes, the value of live linting in the markdown preview tool is much much lower.

KevinMulhern:
Sorry this slipped by me @mao-sz. But I agree, the complexity and work involved to keep things in sync puts this into impractical territory sadly :(

Issue #4524, opened 2024-05-11
Add Feature Flag


bycdiaz:
- [ ] Add feature flag
- [ ] Add placeholder content (content we're hiding)
- [ ] Add test confirming users with flag can observe hidden content
- [ ] Add test confirming users without flag cannot observe hidden content.

## Acceptance Criteria

1. Enable flag for your current user
2. Navigate to x
3. Observe hidden content

JustWaveThings:
From meeting on 5/11:	

- Set up user feature flag check as one first backend tasks and frontend tasks
		- Feature flag conditional:
https://github.com/TheOdinProject/theodinproject/blob/a04874d606f4bae3dddf03b684c14855c66972e8/app/views/static_pages/home.html.erb#L24

PR #4549, opened 2024-05-30
Feature/mvp interview survey

JustWaveThings:
## Because

- We need a feature flag to hide the interview survey feature behind. 

## This PR

- Creates route, controller, view, and updates flipper gates
- Adds feature flag check to controller
- System test coverage for feature flag behavior to ensure no one can access the feature if it is disabled

## Issue

Related to #4524  

~~Missing tests that were called for in that issue.~~ There are now two system tests.

## Additional Information

~~The feature flag does work correctly in the view, but I'm not getting a redirect to work in the controller if the feature is disabled. It still renders the page as if the feature was enabled.~~ This is resolved.

### DIFF
--- a/app/controllers/interview_surveys_controller.rb
+++ b/app/controllers/interview_surveys_controller.rb
@@ -1,0 +1,12 @@
+class InterviewSurveysController < ApplicationController
+  def new
+    unless Feature.enabled?(:survey_feature, current_user)
+      redirect_to dashboard_path, notice: 'Feature not enabled'
+    end
+  end
+
+  def create
+    # puts params[:survey]
+    redirect_to dashboard_path, notice: 'Survey Submitted'
+  end
+end
--- a/app/views/interview_surveys/new.html.erb
+++ b/app/views/interview_surveys/new.html.erb
@@ -1,0 +1,11 @@
+<%= title('Interview Survey') %>
+
+<div>
+  <div class="page-container">
+    <h1 class="page-heading-title">Interview Survey</h1>
+
+    <div class='max-w-xl mx-auto'>
+      <%# where form will be rendered %>
+    </div>
+  </div>
+</div>
--- a/spec/system/interview_survey_spec.rb
+++ b/spec/system/interview_survey_spec.rb
@@ -1,0 +1,28 @@
+require 'rails_helper'
+
+RSpec.describe 'Interview survey' do
+  let(:user) { create(:user) }
+
+  before do
+    sign_in(user)
+  end
+
+  context 'when the feature is enabled' do
+    before do
+      Flipper.enable(:survey_feature)
+    end
+
+    it 'shows the survey' do
+      visit new_interview_survey_path
+      expect(page).to have_content('Interview Survey')
+    end
+  end
+
+  context 'when the feature is disabled' do
+    it 'redirects to the dashboard' do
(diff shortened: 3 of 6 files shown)


app/views/interview_surveys/new.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>InterviewSurveys#new</h1>
+<p>Find me in app/views/interview_surveys/new.html.erb</p>
+<br>
+<% if Feature.enabled?(:survey_feature, current_user) %>
+  <p>My feature is enabled!</p>
+<% end %>
```
bycdiaz, line 6: I'm thinking we should wrap the entire form in the feature enabled check. We wouldn't want a user to see anything in the case of ending up on this page.

Also - this feels like an opportunity for a test. Checking that the form is present when the feature is enabled and checking that the form is not present (e.g. checking for the redirect) when the feature is not enabled.

app/controllers/interview_surveys_controller.rb
```diff
@@ -0,0 +1,10 @@
+class InterviewSurveysController < ApplicationController
+  def show
+    if Feature.enabled?(:survey_feature, current_user)
+      # code to show the survey if not hardcoded into the view
+    else
+      # this does not work
+      redirect_to home_path, notice: 'Feature not enabled'
```
bycdiaz, line 7: Just wanna make sure I'm understanding the comment `# this does not work`. Are we indicating that the logic won't work for the user that tries to get to this and doesn't have the feature enabled?

Or is the comment indicating that the redirect isn't actually working as is?

JustWaveThings: The redirect isn't working as is. If you disable the feature and reload the page, it still loads the page and does not redirect.

I could remove this conditional code from the controller, at least for the MVP, as we can use the feature flag check in the view.

JustWaveThings: I went ahead and removed the conditional code from the controller for now.

app/views/interview_surveys/new.html.erb
JustWaveThings: I'm adding a single input form to show it being wrapped by the flag.

spec/system/interview_survey_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    sign_in(user)
+  end
+
+  context 'when the feature is enabled' do
+    before do
+      Flipper.enable(:survey_feature)
+    end
+
+    it 'shows the survey' do
+      visit 'interview_surveys/new'
```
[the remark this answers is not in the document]
JustWaveThings: Ok. That makes sense. I didn't realize I had to append `new` to the front. I was trying to get `interview_surveys` itself to work.

app/controllers/interview_surveys_controller.rb
```diff
@@ -0,0 +1,12 @@
+class InterviewSurveysController < ApplicationController
+  def new
+    unless Feature.enabled?(:survey_feature, current_user)
+      redirect_to home_path, notice: 'Feature not enabled'
```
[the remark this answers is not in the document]
JustWaveThings: That makes more sense.

KevinMulhern (approved):
Great work! 🚀

PR #4678, opened 2024-07-31
Fix: Add Active Record Encryption keys to test enviroment

Asartea:
## Because
- this is causing tests to fail (see e.g. https://github.com/TheOdinProject/theodinproject/pull/4677), including on a no op

## This PR
- Copies the active record keys from `development.rb` to `test.rb`

## Pull Request Requirements

    - `Fix` - bug fixes


KevinMulhern (approved):
Thanks @Asartea, we have an env var that should be handling this on CI. But it must not be passed to forks.

PR #4743, opened 2024-08-11
Chore / Spike: Upgrade app to rails 7.2

ChargrilledChook:
Quick spike to test effort of ugprading to Rails 7.2, as there aren't many huge changes.

Findings: low effort. 

Notes:
- Linters failing in CI but not locally via `bin/test`, likely because CI uses rubocop binstub and new config adds an explicit `.rubocop.yml` load
- Defaults recommend changing the default max thread count from 5 to 3
- Check if YJIT is in the Docker image we use 
- Some ActiveJob weirdness in console when running app locally, but could just be janky state in local env.
- Several Rails 8 deprecations notices being spammed, particularly for enums - [see change here](https://github.com/rails/rails/commit/8c5425197c7969ff50f675e9792fce1998fb9bc2)

---
 
This commit:
  * Update version in Gemfile
  * Run bundle update
  * Run `app:update` task
  * Fix broken specs
  * Toggle new framework defaults


ChargrilledChook:
Branch is a bit stale and I've lost context, so beginning a fresh branch. In progress, should be low effort, much simpler than than the 7.0  -> 7.1 upgrade

Keeping branch as a reference, will delete once updated branch is good to go

PR #5047, opened 2025-07-13
Chore: Update Rubocop Gems and configuration to use new plugin system

ChargrilledChook:
Because:
Rubocop has recently changed how extensions work to use a plugin system. The changes and related re-structuring of extension Gems were causing a lot of noise in Rubocop output.

This commit:
* Update all flavours of Rubocop Gems
* Remove pessimistic versioning
* Update .rubocop.yml to use the new plugin system
* Set target Ruby and Rails versions
* Add `ExtraDetails` and `DisplayStyleGuide` to AllCops for minor DX
* Explictly add Rubocop Gems that were previously a part of other extensions


Gemfile
```diff
@@ -60,10 +60,13 @@ group :development do
...
-  gem 'rubocop-performance', '~> 1.25', require: false
-  gem 'rubocop-rails', '~> 2.26', require: false
-  gem 'rubocop-rspec', '~> 2.25', require: false
+  gem 'rubocop', require: false
+  gem 'rubocop-capybara', require: false
+  gem 'rubocop-factory_bot', require: false
+  gem 'rubocop-performance', require: false
+  gem 'rubocop-rails', require: false
+  gem 'rubocop-rspec', require: false
+  gem 'rubocop-rspec_rails', require: false
```
ChargrilledChook, line 69: We were requiring some of these (FactoryBot, Capybara, RSpec rails), but they appear to have been extracted to separate Gems as part of the Rubocop plugin changes. Looks like they were previously bundled in as part of the RSpec extension

Gemfile
```diff
@@ -60,10 +60,13 @@ group :development do
   gem 'listen', '~> 3.9'
   gem 'lookbook', '~> 2.3.4'
   gem 'rack-mini-profiler'
-  gem 'rubocop', '~> 1.66', require: false
-  gem 'rubocop-performance', '~> 1.25', require: false
-  gem 'rubocop-rails', '~> 2.26', require: false
-  gem 'rubocop-rspec', '~> 2.25', require: false
+  gem 'rubocop', require: false
+  gem 'rubocop-capybara', require: false
+  gem 'rubocop-factory_bot', require: false
```
ChargrilledChook, line 65: I've removed the pessimistic locking since I don't think it really makes sense for a dev dependency - adds friction to updating without the same safety benefit for functional dependencies.

But I'm happy to add them back if desired

spec/system/admin/announcements_spec.rb
```diff
@@ -63,13 +63,14 @@
...
       expect(page).to have_current_path(admin_announcements_path)
-      expect(page).not_to have_content('Test Message')
+      expect(page).to have_no_content('Test Message')
 
       using_session('learner') do
         sign_in(create(:user))
 
         visit home_path
-        expect(page).not_to have_content('Test Message')
+        expect(page).to have_current_path(home_path)
```
ChargrilledChook, line 72: There's a handful of offences like this I had to fix; I'm assuming because using a positive selector first means you can be sure the page has actually loaded

spec/system/course_lesson_completion_spec.rb
```diff
@@ -34,7 +34,9 @@
     it 'cannot complete a lesson' do
       visit path_course_path(path, course)
 
-      expect(page).not_to have_css('[data-test_id="complete-button"]')
+      expect(page).to have_css('[data-test-id]')
+
+      expect(page).to have_no_css('[data-test-id="complete-button"]')
```
ChargrilledChook, line 39: `data-test_id` was a subtle typo (or it's changed at some point), one of the dangers of using negative selectors.

Fixed while I was here for the new Rubocop offence

.rubocop.yml
```diff
@@ -1,14 +1,16 @@
...
   - rubocop-factory_bot
 
-plugins:
-  - rubocop-performance
-
 AllCops:
+  TargetRubyVersion: 3.4
+  TargetRailsVersion: 7.1
+  ExtraDetails: true
+  DisplayStyleGuide: true
```
ChargrilledChook, line 13: Extra details and style guide link are handy for figuring out what the rule actually wants you to do

PR #5050, opened 2025-07-13
Chore/set cache format version

### LINKED ISSUE #5049: Change format of the Cache Entry
Part of the Rails 7.1 upgrade

```
** Please read carefully, this must be configured in config/application.rb **

Change the format of the cache entry.

Changing this default means that all new cache entries added to the cache
will have a different format that is not supported by Rails 7.0
applications.

Only change this value after your application is fully deployed to Rails 7.1
and you have no plans to rollback.
When you're ready to change format, add this to `config/application.rb` (NOT
this file):
  config.active_support.cache_format_version = 7.1
```

ChargrilledChook:
Closes #5049 

The only risk with this one is we can't downgrade to Rails 7, which _hopefully_ isn't something we've got planned

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -11,6 +11,9 @@ module Theodinproject
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 7.0
 
+    # TODO: Can be removed once config.load_defaults is updated to 7.1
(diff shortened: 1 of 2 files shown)


KevinMulhern (approved):
🎉

PR #5058, opened 2025-07-18
Chore: Bump Sidekiq

KevinMulhern:
Because:
- We were a major version behind.

This commit:
- Also bumps sidekiq-cron - there was a compatibility issue with the new sidekiq dashboard.


ChargrilledChook (approved):
[There's also a fix](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#806) for the Rails compatibility issue in 8.05, but it hasn't been released yet

PR #5060, opened 2025-07-18
Chore: Add an Overlays namespace to components

KevinMulhern:
Because:
- The start of tidying up our components into sensible UI based namespaces or buckets.
- Overlays will contain elements that visually sit on top of main page content.

This commit:
- Moves the flash component from the alerts namespace to overlays
- Moves the modal component into the overlays namespace


ChargrilledChook (approved):
✅

PR #5061, opened 2025-07-19
Chore: Update Active Record encryption configuration

### LINKED ISSUE #5048: Update Active Record Encryption Strategy
Part of the Rails 7.1 upgrade.

```
Active Record Encryption now uses SHA-256 as its hash digest algorithm.

There are 3 scenarios to consider.

1. If you have data encrypted with previous Rails versions, and you have
+config.active_support.key_generator_hash_digest_class+ configured as SHA1 (the default
before Rails 7.0), you need to configure SHA-1 for Active Record Encryption too:
++
Rails.application.config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA1

2. If you have +config.active_support.key_generator_hash_digest_class+ configured as SHA256 (the new default
in 7.0), then you need to configure SHA-256 for Active Record Encryption:
++
Rails.application.config.active_record.encryption.hash_digest_class = OpenSSL::Digest::SHA256

3. If you don't currently have data encrypted with Active Record encryption, you can disable this setting to
configure the default behavior starting 7.1+:
++
Rails.application.config.active_record.encryption.support_sha1_for_non_deterministic_encryption = false
```

ChargrilledChook:
Because:
* Part of the Rails 7.1 upgrade process
* Resolves #5048

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -68,11 +68,9 @@ Rails.application.configure do
   # Highlight code that enqueued background job in logs.
   config.active_job.verbose_enqueue_logs = true
 
-  config.active_record.encryption = {
-    primary_key: 'C697JaYLcwzaSkxtmnedQad2Tl369h4P',
-    deterministic_key: '7Db9oVtlyUn59MkoSnqnwBo17eJqqw7w',
-    key_derivation_salt: 'IlKDRto84iBwxg09w0nbBqlWBe8a2NWT'
-  }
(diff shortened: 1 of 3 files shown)


ChargrilledChook:
Some alternative approaches listed in the ticket #5061

KevinMulhern (approved):
Tested 2fa, and this is working great @ChargrilledChook! :shipit:

PR #5074, opened 2025-07-27
Chore/begin rails 7.2 upgrade

ChargrilledChook:
Because:
We should stay up to date with the latest Rails version.

This commit:
* Run `rails app:update` task and update configuration files
* Add new default health check route
* Bump shoulda matchers to fix incompatibility with Rails 7.2
* Add Brakeman to the project for security scanning (new default)
* Add scripts directory (new default)
* Add temporary framework defaults initializer for Rails 7.2
  * Can be removed after a smooth upgrade
* Add new unsupported browser page
* Remove css compressor setting and Yarn node_modules path
  * Relics of previous asset pipeline setup
* Minor configuration tweaks


ChargrilledChook:
@KevinMulhern I'll tidy this up a bit, squish the commits and add some comments about certain parts, but I think this is 95% of the way there if you want to have a peruse

EDIT: I'll also fix the tests, they were passing locally 😆

ChargrilledChook:
Two test runs, two different types of failures

1st was 

```
ArgumentError: `secret_key_base` for test environment must be a type of String` (ArgumentError)

          raise ArgumentError, "`secret_key_base` for #{Rails.env} environment must be a type of String`"
```

Appears to be some kind of race condition that was introduced from a change
https://github.com/rails/rails/issues/53661
https://github.com/rails/rails/pull/53705

Also seems specific to RSpec parallel and eager loading in CI

Second failure was 
```
  uninitialized constant Rails::Generators
```
and

```
 can't modify frozen Array: ["/home/runner/work/theodinproject/theodinproject/lib", "/home/runner/work/theodinproject/theodinproject/app/adaptors", 
 ```
 
 I'll see if I can figure it out during the week. 
 
 Both appear to be CI specific issues, running tests locally (even with the `bin/ci_spec` command) work fine

**Comment by KevinMulhern** [comment]
Looking good to me @ChargrilledChook.

I'm not overly sure if we'll need or use the pwa and docker files? But don't really have any strong opinions about keeping them around either.

**Comment by ChargrilledChook** [comment]
@KevinMulhern Agreed, I was planning on removing them from this PR although I think the devcontainer is worthwhile adding at some point

The main value being massively reducing the barrier to entry for people setting up the app - it reduces installation steps to: 

1. Install Docker
2. Connect to the devcontainer 

It _almost_ works out of the box, but will require tweaking database.yml a little bit (which probably doesn't hurt since it doesn't look like it's changed much since initial install)

**Comment by KevinMulhern** [comment]
Thanks @ChargrilledChook, I wouldn't be against adding a dev container. We did have it in the past, but it wasn't used much and went unmaintained / fell out of sync with the main app.

But I think it could be different and worth doing now in 2025, with docker basically being ubiquitous.

**Comment by ChargrilledChook** [comment]
@KevinMulhern Maintenance is a fair point. 

[Now](https://guides.rubyonrails.org/7_2_release_notes.html#major-features) that it's officially [supported by Rails ](https://guides.rubyonrails.org/getting_started_with_devcontainer.html) I think the effort goes down a decent chunk, especially since they're maintaining a set of images; the config can be a lot more lightweight and just lean on [the packaged containers](https://github.com/rails/devcontainer)

File: config/routes.rb [hunk 1]
```diff
@@ -4,6 +4,8 @@
   match '422' => 'errors#unprocessable_entity', via: :all
   match '500' => 'errors#internal_server_error', via: :all
 
+  get 'up' => 'rails/health#show', as: :rails_health_check
```
ChargrilledChook, line 7: New default feature / route, don't think Heroku uses it but doesn't hurt to leave it for future compatibility.

Truly a beautiful page

Image

.rubocop.yml
```diff
@@ -220,6 +220,7 @@ Rails/I18nLocaleTexts:
 Metrics/BlockLength:
   Exclude:
     - 'config/environments/*'
+    - 'config/routes.rb'
```
ChargrilledChook, line 223: Unrelated but minor tidy / tweak while touching routes

config/initializers/assets.rb
```diff
@@ -5,10 +5,3 @@
 
 # Add additional assets to the asset load path.
 # Rails.application.config.assets.paths << Emoji.images_path
-# Add Yarn node_modules folder to the asset load path.
-Rails.application.config.assets.paths << Rails.root.join('node_modules')
-
```
ChargrilledChook, line 10: Reasonably sure this is a hangover from old css / asset pipeline. I can't see any assets being referenced straight from node_modules. 

Will check on preview app, let me know if there's a use case I've missed

ChargrilledChook (commented):
Outside of comments, the only other thing worth mentioning is that `config/initializers/filter_parameter_logging.rb` also includes `email` in its filters.

It's probably not a bad idea to include it, but it does remove a little bit of observability.

Happy to go either way.

lib/tasks/heroku.rake
```diff
@@ -1,19 +1,14 @@
...
+    Rake::Task['db:seed'].invoke
+    Rake::Task['curriculum:content:import'].invoke
   end
 
   desc 'Heroku postdeploy task - runs only on review app creation, after release task'
   task postdeploy: :environment do
-    Rake::Task['db:schema:load'].invoke
-    Rake::Task['db:seed'].invoke
-    Rake::Task['curriculum:content:import'].invoke
+    # No op
```
ChargrilledChook, line 12: This gave me grief again like the Rails 7.1 upgrade. I guess they just don't want that API to be public.

I think doing it like this should behave the same way as `db:prepare` is designed to be idemptotent - if the schema exists already it will do nothing.

The only downside is that seeds wil run twice on initial review app creation, but I don't _think_ it's an issue. I couldn't see any issues running the seed command locally multiple times.

db/seeds.rb
```diff
@@ -11,6 +11,7 @@
 load './db/fixtures/lessons/git_lessons.rb'
 load './db/fixtures/lessons/shared_lessons.rb'
 
+require_relative '../lib/generators/rails/seed_uuids/seed_uuids_generator'
```
ChargrilledChook, line 14: Needed to add this due to the eager loading changes.

lib/generators/rails/seed_uuids/seed_uuids_generator.rb
```diff
@@ -1,3 +1,5 @@
+require 'rails/generators'
```
ChargrilledChook, line 1: I think this may have been existing behaviour (or possibly due to eager loading changes), running `db:seed` locally raised an error without this.

lib/tasks/heroku.rake
KevinMulhern: Nice! this looks like a much better way of doing it. 
Running seeds twice should be fine, they've been designed to be idempotent.

It should only affect review apps, and those appear to be working the same, so I think this task can go 💥

KevinMulhern (approved):
This looks great @ChargrilledChook, nice work 💪

PR #5099, opened 2025-08-11
removing_welcome_mail

vivekjain2811:


app/controllers/users/registrations_controller.rb
```diff
@@ -1,12 +1,4 @@
 class Users::RegistrationsController < Devise::RegistrationsController
-  def create
-    super
```
rahul1990gupta, line 3: keep the `create` command. only remove code in `if-else` block

spec/requests/users/registrations_spec.rb
```diff
@@ -38,10 +31,10 @@
         end
       end
 
-      it 'does not send the welcome email' do
+      it 'still creates the user and redirects to dashboard' do
```
KevinMulhern, line 34: Nit: Can we delete this test and the context: `context 'when on a staging environment' do` it's inside please? 

Since we're not making any distinctions between staging and other environments during registration anymore, we don't need this branch of tests.

app/controllers/users/registrations_controller.rb
```diff
@@ -1,10 +1,6 @@
 class Users::RegistrationsController < Devise::RegistrationsController
   def create
```
KevinMulhern, line 2: Nit: Can this controller action be removed please?

It's only delegating to the superclass version of create action now, and we don't need the override anymore.

KevinMulhern (changes requested):
Thanks @vivekjain2811! just a tiny bit of feedback.

spec/requests/users/registrations_spec.rb
vivekjain2811: i have removed this test

vivekjain2811:
@KevinMulhern please review it and tell me if changes required or not i would love to do it.

KevinMulhern:
Hey @vivekjain2811, I'll get this reviewed later today.

KevinMulhern (approved):
This is all working great! thank you for taking the time to do this @vivekjain2811

PR #5140, opened 2025-10-21
Chore: Migrate vertical navigation item component styles to class variants

KevinMulhern:
Because:
- We previously used classy-yaml for style variants, but it is unmaintained and incompatible with the latest versions of ViewComponents.
- Closes https://github.com/TheOdinProject/theodinproject/issues/5098

This commit:
- Migrates vertical navigation item styles to class variants.


KevinMulhern:
Acceptance
- Visit the settings page
- The navigation on the left hand column should look and function the same as production

zachmmeyer (approved):
Confirmed acceptance criteria:
1. Created a new account
2. Visited settings page
3. Compared against production

From the outside it works exactly the same.
✅

PR #5141, opened 2025-10-21
Chore: Migrate theme switcher component styles to class variants

KevinMulhern:
Because:
- We previously used classy-yaml for style variants, but it is unmaintained and incompatible with the latest versions of ViewComponents.
- Closes https://github.com/TheOdinProject/theodinproject/issues/5097

This commit:
- Migrates theme switcher styles to class variants.


KevinMulhern:
Acceptance
- Visit any page
- The theme switcher in the nav bar should look and function the same as production
- Reduce the size of the screen and open the mobile nav menu.
- The theme switcher menu item should look and function the same as production.

PR #5144, opened 2025-10-22
Feature: Center lesson content

KevinMulhern:
Because:
- Previously, to keep the lesson content and sidebar within the width of our page container, which was dictated by the navbar, we positioned lesson content to the left instead of center on the page. This gave us room within the 7xl width to contain both the content and the sidebar.
- But it looked really weird when you got the bottom of the page when completing a lesson, where content was again centered.

This commit:
- Expand the navbar to full width so we no longer have to confine page content within it's width.
- Use a 3 column grid for lesson content - the left column will be empty, the middle column will make sure lesson content is dead center, and the right column will contain the lesson toc sidebar.
- Move the lesson header within the same grid so it can stay in sync with lesson content responsive movements.

### DIFF
--- a/app/views/lessons/_header.html.erb
+++ b/app/views/lessons/_header.html.erb
@@ -1,5 +1,5 @@
-<header class="mb-16 grid grid-cols-12">
-  <div class="flex space-y-4 items-center flex-col col-span-full justify-center xl:space-y-0 xl:space-x-4 xl:flex-row xl:justify-start xl:col-start-2">
+<header class="mb-16">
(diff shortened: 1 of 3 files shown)


KevinMulhern:
Acceptance
- Site navbar is full width and looks decent on all screen sizes.
- Lesson content is centered and looks good on all screen sizes.

zachmmeyer (approved):
This looks great on desktop and mobile views on multiple pages including lessons!

PR #5158, opened 2025-11-08
fix: update github paths in foundation lessons fixture

CommitToday:
## Changes

This PR fixes incorrect GitHub paths in the foundation lessons fixture file. The paths were using kebab-case but should be using snake_case to match the actual file structure in the repository.

## Issue

The GitHub paths in the `foundation_lessons.rb` fixture were inconsistent with the actual file naming convention used in the repository, which was causing issues with content loading and linking.

## Solution

Updated all GitHub paths in the foundation lessons to use snake_case instead of kebab-case, aligning them with the actual file names in the repository.

## Files Changed

- `db/fixtures/lessons/foundation_lessons.rb` - Updated GitHub paths from kebab-case to snake_case


mao-sz:
Incomplete duplicate of #5157

PR #5159, opened 2025-11-08
Chore/upgrade to rails 8

ChargrilledChook:
**Because**
We should stay up to date on Rails versions

**This commit**
* Run rails update task + align new configs
* Set framework defaults to 8
* Update env and config files


KevinMulhern (approved):
LGTM @ChargrilledChook, 🚢

PR #5161, opened 2025-11-08
Chore: Add ruby-lsp-rspec Gem

ChargrilledChook:
## Because
Offers several nice conveniences for running specs locally, such as tracking timings and 1 click IDE debugging

## Examples

Integration with the VSCode test runner

image

VSCode debugging

image

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -290,7 +290,6 @@ GEM
       mime-types-data (~> 3.2015)
     mime-types-data (3.2024.0806)
     mini_mime (1.1.5)
-    mini_portile2 (2.8.9)
     minitest (5.26.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -310,8 +309,7 @@ GEM
     netrc (0.11.0)
     newrelic_rpm (9.16.0)
     nio4r (2.7.3)
-    nokogiri (1.18.10)
-      mini_portile2 (~> 2.8.2)
+    nokogiri (1.18.10-arm64-darwin)
```
ChargrilledChook: Something funky happened when adding the gem, it shouldn't be locking to a platform 🤔

PR #5164, opened 2025-11-13
Chore: Refactor controller feat flag gates

JoshDevHub:
## Because

The current checks for `Flipper` need to be repeated across any controller action that requires a feature flag check. There are a couple of compelling reasons to DRY this up:
1. It requires more work than necessary to remove the feature flag checks from a controller as you have to go and edit each action in the controller--and that's also going to be more error prone than a potentially simple one-line edit.
2. If you want to do something like change where the redirect goes when the feature is disabled, you again have to edit every place `Flipper` is checked.

I also just think there are benefits to establishing a more Rails-y convention on checking feature flags in the controller.

## This PR

- Extracts a concern that wraps the `Flipper` check
- `extend` the concern into `ApplicationController`
- Refactor the `InterviewSurveyController` to use the new concern

## Issue

Closes #XXXXX

## Additional Information

This is pretty shamelessly stolen from https://kaspth.com/posts/clear-up-your-rails-controllers-with-before_action-wrappers

The article argues for a different redirect when a feature is disabled, but I just kept what was there previously. Not sure it matters a ton in our context. Anything that tries to be opaque about what the user is accessing probably doesn't have much value with a fully open source app. Willing to change it though!

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/app/controllers/concerns/requires_feature.rb
+++ b/app/controllers/concerns/requires_feature.rb
@@ -2,7 +2,7 @@ module RequiresFeature
   def requires_feature(name, **)
     before_action(
       lambda do
-        return if Flipper.enabled?(name, current_user)
+        return if Feature.enabled?(name, current_user)
 
         redirect_to dashboard_path, notice: 'Feature not enabled'
       end,


KevinMulhern (approved):
Sorry Josh, I gave this another pass and noticed the Feature/Flipper bit.

I realised I never properly surfaced some updates I made to our feature flag setup last year. I did document it, but I'm just now seeing the docs ended up buried in the repo’s wiki without a link on the front page 🙈
https://github.com/TheOdinProject/theodinproject/wiki/Feature-Flags

Quick TL;DR:
We have a feature-flags task that runs during seeding. If we add our feature flag names to that list, it will:
- Automatically create the flag in Flipper so we don’t need to set it up manually in each environment.
- Automatically enable those flags in dev and review-app environments, which makes QA much smoother (otherwise we have to go into Admin → Flipper and toggle them by hand).

It would be worth adding the `:survey_feature` to it to help streamline things for this feature.

KevinMulhern (approved):
🚀 🚢

PR #5167, opened 2025-11-17
Feature: Group lessons by path in Installation Guides page

### LINKED ISSUE #5166: Feature Request: Add Node "Installing PostgreSQL" lesson to Installations Guide page
### Checks

- [x] The title of this issue follows the `Feature Request: brief description of feature request` format, e.g. `Feature Request: Add a dark mode to the website`

As far as I can see, the only installations lesson not marked as `installation_lesson: true` is the [Node course's "Installing PostgreSQL"](https://github.com/TheOdinProject/theodinproject/blob/82f0e502e5e10ff63def3c274470991d6f12280c/db/fixtures/lessons/node_js_lessons.rb#L84).

If we just mark it as an installation lesson, the [Installation Guides page](https://www.theodinproject.com/guides/installations) will have two entries for "Installing PostgreSQL" with not quick way to differentiate between them.

#### Simple solution

We could change the names of both lessons to "Installing PostgreSQL (Ruby on Rails)" and "Installing PostgreSQL (Node)" which will solve the Installation Guides page problem, but then:
- The lesson names might seem odd when you look only at the course contents, if someone isn't aware about the Installation Guides page.
- The lesson slugs will change (shouldn't be an issue in the curriculum since a grep returns no current links to either Postgres installation lesson anywhere).

#### Fancier solution

Alternatively, we could add a `disambiguate_by_course` boolean to those lesson's fixture hashes, add the appropriate column, then do something like the following in the lesson model:
```rb
  def display_title(for_guide_page = false)
    title_to_display = is_project ? "Project: #{title}" : title

    if for_guide_page && disambiguate_by_course
      "#{title_to_display} (#{course.title})"
    else
      title_to_display
    end
  end
```

That should leave the lessons as is except for the Guide page where only the "Installing PostgreSQL" lessons will have the course title tacked on.

### Acceptance criteria

- [ ] In the Installation Guides page, both pathways' "Installing PostgreSQL" lessons are listed and clearly disambiguated.

### Additional Comments

May be a simple best of both worlds solution I haven't considered?

mao-sz:
## Because

The [installation guides](https://www.theodinproject.com/guides/installations) page currently does not have the Node course's "Installing PostgreSQL" lesson. Since that lesson shares its title with the Rails course's one, we need to disambiguate between them and grouping the lessons by pathway seems most sensible (no need to mess with the DB or affect lesson entries outside of the Installation Guides page).

## This PR

- Replaces the section partial with a more generic LessonList component
  - Tests added/amended accordingly
- Marks Node's "Installing PostgreSQL" lesson as an installation lesson
- Groups installation guide page lessons by path

## Issue

Closes #5166

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/spec/components/lesson_group_component_spec.rb
+++ b/spec/components/lesson_group_component_spec.rb
@@ -18,7 +18,6 @@ RSpec.describe LessonGroupComponent, type: :component do
 
     render_inline(component)
 
-    expect(page).to have_css('div[data-test-id="lesson-list"]')
     expect(page).to have_content('Lesson title', count:)
   end
 end


spec/components/lesson_list_component_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(page).to have_css('h3', text: 'HTML Basics')
+  end
+
+  it 'renders the list of lessons' do
+    component = described_class.new(title: 'HTML Basics', lessons: [], current_user: nil)
+
+    render_inline(component)
+
+    expect(page).to have_css('div[data-test-id="lesson-list"]')
+  end
```
mao-sz: Ideally, I'd want this to pass a mock `lessons` array then I can confirm it does indeed render the expected content, rather than relying on just the container test-id.

mao-sz: Edit: sorted in https://github.com/TheOdinProject/theodinproject/pull/5167/commits/82a895894607948ee9fa8d3bcebbbe52d38dec0d (cheers, Kev)

app/views/guides/installations/index.html.erb
```diff
@@ -2,20 +2,9 @@
...
-                <p class="text-gray-600 dark:text-gray-300">
-                  <%= lesson.display_title %>
-                </p>
-            </div>
-          <% end %>
-        <% end %>
-      <% end %>
+  <div class="max-w-xl mx-auto space-y-5">
+    <% @lessons_by_path.each do |path, lessons| %>
+      <%= render LessonListComponent.new(title: path.title, lessons:, current_user: nil) %>
```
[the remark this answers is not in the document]
JoshDevHub: The way you have it is fine as far as I know.

I will say I'd be a bit inclined to have `current_user` here be a default parameter so you're not having to explicitly pass `nil`, which just feels strange to me.
```rb
# back in the component file
def initialize(title:, lessons:, current_user: nil)
```

app/views/guides/installations/index.html.erb
```diff
@@ -2,20 +2,9 @@
...
     <% end %>
-  </section>
+  </div>
```
JoshDevHub: Are you intentionally changing the semantics of the HTML here? Not saying you shouldn't, just idly wondering what the reason is.

JoshDevHub (commented):
A couple of ultra nits.

Any chance you can upload a pic of what the installation page looks like? I'm feeling too lazy to pull this down and look 😅

mao-sz:
For some reason I'm unable to create a review app on Heroku (clicking "create review app" just loads for a bit then does nothing (console just logs a 404 for whatever endpoint that leads to) despite settings suggesting I should have the perms for it :shrug:). Here's a screencap of the installation guides page and in-course section list.

https://github.com/user-attachments/assets/e84ae4cd-95e9-490e-8b93-5c1108a4ae08

JoshDevHub:
> For some reason I'm unable to create a review app on Heroku (clicking "create review app" just loads for a bit then does nothing (console just logs a 404 for whatever endpoint that leads to) despite settings suggesting I should have the perms for it 🤷). Here's a screencap of the installation guides page and in-course section list.

Review apps don't work right from PRs where the origin is a fork (which I'm pretty sure is the problem you have here), and it's apparently kinda tricky to configure to where it would work.

Since you're a maintainer though, you should be able to push branches here from a direct clone of this repo, and that should allow you to create Review Apps. This is what I do anytime I'm making something where I want a Review App (at least if I remember to do so 😅 )--and Kev is fine with us doing that.

Don't feel like you  have to remake this PR or anything for it though. Pretty sure Kev can come in here and spin one up using his elevated perms. Just something to note for the future.

Edit: looks really nice by the way!

ManonLef:
It has been a while since I have used it but I always thought the tag should do the trick. I guess not

mao-sz:
Ah, I didn't realise review apps from forks required elevated perms.

app/components/lesson_list_component.rb
```diff
@@ -0,0 +1,11 @@
+class LessonListComponent < ApplicationComponent
```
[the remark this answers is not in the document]
mao-sz: Oh actually that's a good point. I think I started with `LessonList` when I was looking only at the list itself but then never renamed when I looked at the whole group.

spec/components/lesson_group_component_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  it 'renders the list of lessons' do
+    count = rand(10)
+    component = described_class.new(
+      title: 'HTML Basics',
+      lessons: create_list(:lesson, count, title: 'Lesson title'),
+    )
+
+    render_inline(component)
+
+    expect(page).to have_css('div[data-test-id="lesson-list"]')
```
KevinMulhern: Nit: I'd be tempted to drop this expectation. The other expectation tests the behaviour perfectly, and should should cover us better if the component wasn't rendering the lesson list during because of a refactor or some other reason.

KevinMulhern (approved):
Nice work @mao-sz, this is looking great! components ftw 💪

PR #5169, opened 2025-11-19
Chore: Make coverage reports opt-in

JoshDevHub:
## Because

There are a couple reasons I thought this was a good idea:
1. Coverage reports take up a not-insignificant amount of time in a test run. On the machine I'm making this PR for, generating the reports consumes over 10 seconds in a full test run.
  **Without Coverage**image
  **With Coverage**image
2. In a partial run, the generated coverage reports aren't very useful. I often run one file or sometimes even one spec. In such a run, the coverage is more noise than anything, as it doesn't have the full context of the tests that weren't run in order to generate an accurate report. This is also the case in the current CI config, where tests are executed in parallel on different workers, meaning none of the coverage reports generate an accurate picture.

## This PR

- Wraps require and init for `SimpleCov` in a conditional check that looks for a `'COVERAGE'` env variable. To generate a report, the dev now needs to run something like `COVERAGE=1 bin/rails spec`

## Issue

Closes #XXXXX

## Additional Information

If we decide we want it, there are ways to have SimpleCov emit JSON reports and have it merge them together. This could potentially be used to generate an accurate, combined report in the CI env. I figure that'd be something for a different PR/issue though.

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,6 +1,8 @@
-require 'simplecov'
+if ENV.fetch('COVERAGE', false)
+  require 'simplecov'
+  SimpleCov.start 'rails'
+end
 
-SimpleCov.start 'rails'
 ENV['RAILS_ENV'] ||= 'test'
 
 require File.expand_path('../config/environment', __dir__)


KevinMulhern (approved):
Nice work @JoshDevHub, thats a big chunk of time saved!

I think your idea of emitting JSON reports via CI sounds like it could be really neat.
Although, I'd be kind of tempted to remove SimpleCov altogether. It was useful when we needed better coverage with system tests a few years ago. But I can't remember the last time we've looked at it since then.

But, if you and others on the team find it useful we should definitely keep it. Just wanted to put it out there lol.

PR #5170, opened 2025-11-20
Chore: Reconfigure `debug` gem in Gemfile

JoshDevHub:
## Because

It speeds up runs of the test suite by a healthy chunk to not have the debugger loaded. This is because it periodically blocks the main thread while it does its thing--even when no breakpoint is set and it's not being actively used by the developer. Loading the debugger every time into the test environment can be prevented by setting `require: false` on the gem in the `Gemfile`.

On my machine (which I admit is a bit older and slower than the average silicon Mac), it averages about 7-8 seconds saved over a full run to not load the debugger:
**With debug loaded**image

**Without debug loaded**image

While making this change, I noticed that the `debug` gem is listed at the top level of Gemfile, and therefore theoretically included in the production bundle. This is a big smell to me for both security and performance reasons. So I also figured I'd move it to the local group (development and test).

## This PR

- Moves the `debug` gem to the group for development and test environments
- Sets `require: false` on the gem so it's not always loaded

## Issue

Closes #XXXXX

## Additional Information

Sometimes you *do* want to debug. With this change, the `debug` gem is no longer required throughout the app. If you're wanting to debug in development, you'd have to explicitly require it wherever you're wanting to use it
```rb
# app/controllers/lessons_controller.rb

require 'debug' # now you need this

  def show
    binding.b # or this breakpoint won't work
  end
end
```
For debugging tests, the developer can either do the same thing (`require` the gem into the file they're wanting to set a breakpoint at) or RSpec can be ran with the `-rdebug` flag, which will load the gem.

Personally, I think this is a very minor inconvenience. Also glad to document these somewhere (the wiki?) if people think it'd be helpful.

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,7 +71,7 @@ end
 
 group :development, :test do
   gem 'brakeman', require: false
-  gem 'debug', '~> 1.10', require: false
+  gem 'debug', '~> 1.10', require: 'debug/prelude'
   gem 'dotenv-rails', '~> 3.1'
   gem 'parallel_tests', '~> 5.4'
   gem 'rspec-rails', '~> 7.0'


KevinMulhern (approved):
Nice catch! loving these performance improvements!

ChargrilledChook:
@JoshDevHub an alternative that (I think) has the same performance benefit without needing to manually require the gem is `debug/prelude` - essentially a way to lazy load the Gem if its invoked

From one of the Gem maintainers: 

https://st0012.dev/my-ruby-debugging-tips-in-2025

I noticed a decent performance improvement switching to this at work in two different apps

---

(Relevant snippet from the link)

Use `gem "debug", require: "debug/prelude"` in your Gemfile instead.

debug.gem is activated when required, which usually isn't necessary when you're not debugging.

By requiring debug/prelude, it defines the breakpoint methods like breakpoint and binding.break, but doesn't activate the gem immediately.

(This has been the default in newly generated Rails projects since Rails 7.2.)

JoshDevHub:
> Use `gem "debug", require: "debug/prelude"` in your Gemfile instead.
> 
> debug.gem is activated when required, which usually isn't necessary when you're not debugging.
> 
> By requiring debug/prelude, it defines the breakpoint methods like breakpoint and binding.break, but doesn't activate the gem immediately.
> 
> (This has been the default in newly generated Rails projects since Rails 7.2.)

Oh I didn't know about this. That's definitely better. I tested locally and has the same perf benefits as `require: false`. I'll push  a commit up with that change in it.

ChargrilledChook (approved):

PR #5193, opened 2025-12-11
Fix: Lint all JS/CSS files with Standard/Stylelint

mao-sz:
## Because

The current `lint` script runs only in `app/javascript` which therefore misses the JS config files in the repo root. Can't see a reason for these to not be linted either.

The current `stylelint` script also misses `application.tailwind.css`.

## This PR

- Amends paths for `lint` and `stylelint` scripts
  - Removes `--verbose` flag for `standard` as it isn't a valid flag.
- Amends stylelint config with Tailwind at-rules and `import-notation` (for consistency within file)
- Fixes JS and CSS lint errors in missed files.

## Issue

Closes #XXXXX

## Additional Information

Opening up the `standard` path revealed JS controllers in
- https://github.com/TheOdinProject/theodinproject/blob/main/app/components/project_submissions/item_controller.js
- https://github.com/TheOdinProject/theodinproject/blob/main/app/components/overlays/dialog_controller.js

Is this intentional? I thought they would go in `app/javascript/controllers`. 

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/app/components/project_submissions/item_controller.js
+++ b/app/components/project_submissions/item_controller.js
@@ -1,25 +1,25 @@
-import { Controller } from '@hotwired/stimulus';
+import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
-  static targets = ['likeCount'];
+  static targets = ['likeCount']
 
   static values = {
-    likeCount: String,
-  };
+    likeCount: String
+  }
 
-  likeCountTargetConnected(element) {
-    const currentLikeCount = element.innerText;
+  likeCountTargetConnected (element) {
+    const currentLikeCount = element.innerText
 
     if (currentLikeCount !== this.likeCountValue) {
-      this.likeCountValue = currentLikeCount;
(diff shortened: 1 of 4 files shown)


mao-sz:
Cheers @KevinMulhern 
Thoughts on the 2 JS controllers bit in the additional info part of the PR form? If action is required then I can PR separately

KevinMulhern:
Sorry I missed that @mao-sz. Yep, those are intentional.

Ideally we’d only have generic controllers that can be used anywhere in the app, living in `app/javascript/controllers`. In practice though, that’s easier said than done, and I think we have a few controllers there with fairly narrow use cases.

View components make it easier though. They allow Stimulus controllers to live in a component’s sidecar when the behavior is tightly coupled to that component. It's mostly a DX win - everything related to the component lives in one place, so there’s less context switching - but it also clearly communicates that the controller is not meant to be reused elsewhere.

mao-sz:
Gotcha :+1:

PR #5221, opened 2026-01-29
docs: clarify documentation-only and minor contribution workflow

ahmasmibhanu:
This PR adds a new section to the web app contributing guide clarifying that
documentation-only and minor content changes do not require running the full
local development environment.

The goal is to reduce friction for new contributors while keeping expectations
clear about when local setup and testing are required.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,7 @@ This contributing guide assumes you have followed the instructions in our genera
 ## Table of Contents
 
 - [How to Contribute](#how-to-contribute)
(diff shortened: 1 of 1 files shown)


JoshDevHub:
Make an issue before putting in a PR with this much in it.

And some advice if you do make an issue and actually care about contributing and aren't just shotgunning similar doc PRs to dozens of different projects: try to not use AI. It's much much much more obvious than you think it is. I could be interested with accepting a change like this, but it needs to be about 10x less verbose and mealy mouthed. No one likes reading slop. Be straight and to the point.

PR #5278, opened 2026-04-06
Feature: Improve accessible text for links wrapping images in lessons

mao-sz:
## Because

Currently, images in lessons are wrapped in anchors - the accessible text for these anchors are just the images' alt text.

For example, in [Understanding Errors](https://www.theodinproject.com/lessons/foundations-understanding-errors#the-anatomy-of-an-error), for the first image when tabbing through and navigating by link, Orca reads out "reference error example - link" and that's it. It's not 100% clear that this link's purpose is associated with opening an image; it could be a link to an article or code sandbox.

A lot of our lessons' images' alt text can do with improving, but it would make sense to ensure that these associated anchors have a more specific accessible name at the conversion time. Improving the alt texts is a separate and significantly larger task.

After chatting with Eric about approaches, giving the anchors an aria-label with `image showing <alt text>` should be sufficient. The anchors have a clearer purpose while the images are left alone to avoid redundancy.
Orca reads the link as "image showing reference error example - link" and reads the image as "reference error example - image".

## This PR

- Adds aria-label to anchors that wrap images in the markdown converter
- Amends associated test

## Issue

N/A

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/lib/kramdown/converter/odin_html.rb
+++ b/lib/kramdown/converter/odin_html.rb
@@ -12,7 +12,7 @@ module Kramdown
 
         attributes = { href: element.attr['src'], 'aria-label': "image showing #{element.attr['alt']}" }
           .merge(EXTERNAL_LINK_ATTRIBUTES)
-        %(<a#{html_attributes(attributes)}>#{super}</a>)
+        "<a#{html_attributes(attributes)}>#{super}</a>"
       end
 
       def convert_a(element, indent)
--- a/lib/kramdown/converter/odin_html.rb
+++ b/lib/kramdown/converter/odin_html.rb
@@ -10,7 +10,8 @@ module Kramdown
       def convert_img(element, _indent)
         return super if @stack.last.type == :a || element.attr['alt'] == ''
 
-        attributes = { href: element.attr['src'] }.merge(EXTERNAL_LINK_ATTRIBUTES)
+        attributes = { href: element.attr['src'], 'aria-label': "image showing #{element.attr['alt']}" }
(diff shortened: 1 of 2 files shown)


lib/kramdown/converter/odin_html.rb
```diff
@@ -10,7 +10,8 @@ module OdinHtml
       def convert_img(element, _indent)
         return super if @stack.last.type == :a || element.attr['alt'] == ''
 
-        attributes = { href: element.attr['src'] }.merge(EXTERNAL_LINK_ATTRIBUTES)
+        attributes = { href: element.attr['src'], 'aria-label': "image showing #{element.attr['alt']}" }
+          .merge(EXTERNAL_LINK_ATTRIBUTES)
         %(<a#{html_attributes(attributes)}>#{super}</a>)
```
mao-sz: Is there a reason this uses `%()` instead of `"`? A quick grep shows this is the only case where we use `%` in this way, and I can't see why it shouldn't just be `"` for consistency since there aren't any double quotes in the string itself.
```suggestion
        "<a#{html_attributes(attributes)}>#{super}</a>"
```

KevinMulhern: Nice catch! I think we might have been anticipating that we could have double quotes strings within this in the future and used the percent notation to auto-escape those (`%(<a class="external-link" #{html_attributes(attributes)}>#{super}</a>)`). 

But, YAGNI - switching to double quotes will be fine.

KevinMulhern (approved):
LGTM 🚀

PR #5289, opened 2026-04-14
Feature: Differentiate between links that open in a new tab vs same tab

mao-sz:
## Because

With #5247 now merged, we no longer have a clear difference between links that open in a new tab vs same tab on left click. A typical "new tab" icon on such links would be sensible.

## This PR

- Create SVG icon for "open in new tab"
- Create NewTabLinkComponent
- Use NewTabLinkComponent for links that make sense to open in a new tab by default
- Create/amend tests as necessary

## Issue

N/A

## Additional Information

- Link/icon accessible text amended to follow https://www.w3.org/WAI/tutorials/images/functional/#example-3-icon-image-conveying-information-within-link-text. Orca reads out the new lesson changelog link as "see lesson changelog new tab link". We didn't really need the title as well - the tooltip would only show if hovering over the icon anyway
- Icon SVG obtained from https://www.svgrepo.com/svg/471618/link-external-02 - no attribution required due to CC0 license
- I got rid of the unique leading icons for the lesson edit/issue links because they looked super strange having both leading and trailing icons:
  
  image

- The two contribution card links I amended to avoid having two different links with the same text label.
  
## Pull Request Requirements

    - `Fix` - bug fixes


app/views/static_pages/support_us.html.erb
```diff
@@ -15,7 +15,12 @@
     </div>
 
     <div class="mt-8 text-center">
-      <%= link_to 'Donate now!', 'https://opencollective.com/theodinproject/donate?amount=5', class: 'button button--primary text-base px-12 py-4', target: '_blank' %>
+      <%= render NewTabLinkComponent.new(
```
KevinMulhern, line 18: I'd be templated to remove opening in a new tab from the donate button. It _shouldn't_ take the user out of their flow like opening the chat side-by-side with lesson content, or comparing solutions would.

app/views/static_pages/contributing/_contributing_options.html.erb
```diff
@@ -5,20 +5,18 @@
     <% contributions.each do |contribution| %>
       <%= render CardComponent.new(classes: 'flex flex-col flex-between h-full border-t-8 border-t-gold-600 dark:border-t-gold-700') do |card| %>
         <% card.with_header do %>
-          <h3 class="text-xl font-medium text-center"><%= contribution[:title] %></h3>
+          <h3 class="text-xl font-medium text-center">Contribute to the <%= contribution[:title] %></h3>
```
KevinMulhern, line 8: Good spot with this! we might be able to simplify the copy on this page more by dropping the "Contribute to the" here since we're already mentioning contributing in the two titles above the cards.

Image

app/views/lessons/show.html.erb
```diff
@@ -16,26 +16,29 @@
...
-                <%= link_to github_report_url(@lesson, 'broken_link'), target: :_blank, rel: 'noreferrer', class: 'inline-flex items-center underline hover:no-underline text-gray-600 hover:text-gray-800 dark:text-gray-400 dark:hover:text-gray-300 pr-5 text-sm' do %>
-                  <%= inline_svg_tag 'icons/link.svg', class: 'h-4 w-4 mr-1 inline', aria: true, title: 'Report broken link', desc: 'Broken link icon' %>
-                  <span>Report broken link</span>
-                <% end %>
-
-                <%= link_to github_report_url(@lesson, 'suggestion'), target: :_blank, rel: 'noreferrer', class: 'inline-flex items-center underline hover:no-underline text-gray-600 hover:text-gray-800 dark:text-gray-400 dark:hover:text-gray-300 pr-5 text-sm' do %>
-                  <%= inline_svg_tag 'icons/flag.svg', class: 'h-4 w-4 mr-1 inline', aria: true, title: 'Report other issue', desc: 'Report icon' %>
-                  <span>Report other issue</span>
-                <% end %>
+                <%= render NewTabLinkComponent.new(
```
KevinMulhern, line 19: I think our text links are where this pattern definitley fits best rather than cta-like buttons!
But since there quite a few links here in one place, do we lose a bit of 'scannability' by swapping the specific icons (like the GitHub logo and flag) for a generic new-tab icon? 🤔

KevinMulhern (commented):
Nice work on the component! It looks _really_ good. 

I think this absolutely makes sense and is a UX QOL upgrade for inline links and links that have an ambiguous destination. But I’m wondering if we might want to be more selective with its scope. Since buttons like 'Open Discord', 'Live preview', 'View on GitHub' already tell the user they’re heading elsewhere, we might be able to skip the icon in those spots. I think having the icon on those could feel a bit like visual noise rather than a really useful signal. What do you think about only using it for links where the destination is more ambiguous?

app/views/static_pages/support_us.html.erb
mao-sz: I think this makes sense

app/views/lessons/show.html.erb
mao-sz: I'm not so sure we should drop the new tab icons in favour of using other icons to differentiate between the links. Those text links are in pages with lots of other same-tab text links, so I think we really ought to make it visually clear as well that they open in new tabs. Similar to MDN docs:
image


I think for me, any reduction in scannability would be to do with all links being roughly equal length (which blends the "Report" bits together). If we only had 2 links there (no dedicated broken links one) then it wouldn't be too much of an issue, I don't think.

Proposal:

image

1. Most doc pages with an edit link use the term "edit" since that's what it is. I feel "improve" leads to a lot of go-getters directly PRing any kind of "improvement", both superficial and large scale.
2. Issues aren't strictly for reporting bugs. While broader less-actionable ideas tend to get moved to Discussions, most suggestions are actionable and deserving of an issue (and the issue template is called `suggestion.yml` after all). 

To my eyes, those links scan much more easily than even 3 equally long ones with differentiating icons. What do you think?

mao-sz:
I think you make a good point about text links vs "button" links, though I'm a little uncertain going back through WCAG.

The criteria I'm trying to meet is [Change on Request (SC 3.2.5)](https://www.w3.org/WAI/WCAG22/Understanding/change-on-request). One of the recommended techniques for meeting it is [H83](https://www.w3.org/WAI/WCAG22/Techniques/html/H83) which indicates a change of context in the link itself. I realised I misinterpreted this as *the* success criteria and not just a recommended technique. However, [F22](https://www.w3.org/WAI/WCAG22/Techniques/failures/F22) is a specific failure criteria for SC 3.2.5 where one of the examples is:
> A user clicks on a link, and a new window appears. The original link has no associated text saying that it will open a new window. 

Though the F22 description does also say 

> This is fine when the user has interacted with a piece of User Interface and expects to get a new window, such as an options dialogue. The failure comes when pop-ups appear unexpectedly. 

So I guess as long as the button links we have have text that's much clearer about the destination, we *could* omit the icon, but an icon would definitely unambiguously meet the criteria for both visual and AT users.

We ideally shouldn't have two button links side by side where one opens in a new tab and the other in the same tab, at least without accessible distinction.

Proposal:

1. Rename component to `NewTabTextLinkComponent`
1. Revert button links to non-component
1. Rename "View code" to "View GitHub repo" - unambiguous about location, more justifiable change of context ("View code" could be interpreted as some inline embed). Also allows it to be side-by-side with "Live preview" which also changes context
1. Make "Donate" button links open in same tab

mao-sz:
Add-on thought:

Do we even really need to make any of our button links open in a new context? The more I think about it, the less I can justify any of them. I don't see any disruptive behaviour with them opening in the same tab and users can still choose to open them in a new tab/window if they desire.

e.g. if you're viewing community submissions then opening a repo or live preview in the same context won't disrupt any flow of anything. Just tested opening a live preview in the same tab and hitting "Back" retained the submissions page I was on (not first page), so that is a non-issue.

The only ones that really make sense to me to open in a new tab are:
1. Lesson footer report/edit links since people will often want to do so while referencing the live page simultaneously
1. "Open in Discord" since that either opens Discord in the browser or is a redirect link to the app. So it makes sense to open it in a new tab IMO.

That also lessens the burden over thinking about visual noise, ambiguous wording, text vs CTA etc. We retain `NewTabLinkComponent` and just use it for those two use cases, and all other links are same tab by default so don't need the component. The Discord button would be the only button link with the icon, so that shouldn't be much visual noise with everything else.

Asartea:
> e.g. if you're viewing community submissions then opening a repo or live preview in the same context won't disrupt any flow of anything. Just tested opening a live preview in the same tab and hitting "Back" retained the submissions page I was on (not first page), so that is a non-issue.

I disagree with this: a lot of the time I'm looking at submissions I want to look at more than one at a time, or even just at both the repo and the live preview, and having the link open in a new tab means I can just click both buttons, rather than clicking one, looking at it, going back, and then clicking the other

mao-sz:
We're not removing the ability to do that though, since you can still middle-click or use the context menu. I cannot see a strong reason to force those links to have to open in a new tab every time - we should allow the user to decide.

Whereas with the edit/report links in lessons, my justification for having them always open in a new tab is because all if those typically involve processes with multiple pages and at least one form in the middle. So navigating back to the lesson would require more than one step in most cases and I have no idea how GitHub handles navigating back through submitted forms/editors but that just feels unnecessary to me. I think it makes clear sense to have the edit/report stuff open separately, then once it's done, people can exit out of it and return to the lesson in a single action without any chance of going back through submitted forms.

And while explicit warning about new tab opening isn't hard required for conformity to SC 3.2.5 but an advisory technique ([G201](https://www.w3.org/WAI/WCAG22/Techniques/general/G201)), I don't see why we need to deal with making some new-tab links explicit about it and some not, and you'd have to guess on context (which isn't consistent either - we aren't defaulting to "internal same tab - external new tab").

app/views/lessons/show.html.erb
KevinMulhern: I think this is all fair! I'm digging those copy changes, much better 🤌

KevinMulhern:
I think the submission links would be best remaining opening in a new tab for similar reasons as the edit/report links. It would be okay for some of the earlier projects, but the middle and later projects could be many pages deep.

Asartea's comparison workflow is a good point too, many learners open a few to compare the preview/code. Using the middle-click button is an okay comprimise, but is opening in new tab the better default? whatever causes the least friction with the workflows users have with submissions can be the guide.

> The criteria I'm trying to meet is [Change on Request (SC 3.2.5)](https://www.w3.org/WAI/WCAG22/Understanding/change-on-request).

Side note, but definitely worth mentioning. I noticed this is an AAA rule. Most sites (including us 😛) will aim for AA because of how constraining and dogmatic AAA can be. But I think it's a noble goal to aim for it where it's reasonable. This one is probably a good example of that, it doesn't seem that difficult or constraining for us to get most of the way there minus a couple of trade-offs.

I think the proposal you wrote up here is the way ™️ 
> 1. Rename component to NewTabTextLinkComponent
> 2. Revert button links to non-component
> 3. Rename "View code" to "View GitHub repo" - unambiguous about location, more justifiable change of context ("View code" could be interpreted as some inline embed). Also allows it to be side-by-side with "Live preview" which also changes context
> 4. Make "Donate" button links open in same tab

mao-sz:
While it is AAA and so less of a concern than AA (for strict conformance), I think if there is a reasonable path to meeting individual AAAs then they're still worth exploring, even if a site doesn't meet all or even just a good portion of AAA criteria. But I'll concede on the project submission button links, the alternative is ultimately still practical. I'll get on these changes when I have the time

KevinMulhern:
> While it is AAA and so less of a concern than AA (for strict conformance), I think if there is a reasonable path to meeting individual AAAs then they're still worth exploring, even if a site doesn't meet all or even just a good portion of AAA criteria.

Agreed! thank you for exploring it and making these changes. Users will be better off for it.

I remember one time we wanted to meet AAA contrast. That was a nightmare. It was way too constraining and almost every small styling change got bogged down with multiple tweaking iterations to make dark mode look decent and not like the Las Vegas strip at night! lol

app/views/static_pages/support_us.html.erb
```diff
@@ -15,14 +15,8 @@
...
     <div class="mt-8 text-center">
-      <%= render NewTabLinkComponent.new(
-            text: 'Donate now!',
-            href: 'https://opencollective.com/theodinproject/donate?amount=5',
-            classes: 'button button--primary text-base px-12 py-4',
-            noreferrer: false
-          ) %>
+      <%= link_to 'Donate now!', 'https://opencollective.com/theodinproject/donate?amount=5', class: 'button button--primary text-base px-12 py-4', target: '_blank' %>
     </div>
-    <%# <script src='https://opencollective.com/theodinproject/banner.js?style={"h2":{"color":"rgb(112, 112, 128)"}}'></script> %>
```
mao-sz, line 25: This commented-out script can be removed, right? I imagine if we needed it, it wouldn't be commented out and we'd have probably noticed a while back already.

KevinMulhern: Yep, that can go! 💥

KevinMulhern (approved):
LGTM, nice work @mao-sz 💪

PR #5301, opened 2026-04-23
Fix: Subtract automatic project likes from daily stats

### LINKED ISSUE #5300: Bug: Daily Summary notification service is including automatic likes in Project Like Count
### Checks

### Bug description

The daily metrics posted to Discord are primarily intended to portray member activity. Since the single automatic project like granted upon project submission is performed by the system and not a member, it should be omitted from the count of projects liked.

### How to reproduce

1. Open up The Odin Project's Discord server.
2. Review recent posts in the top-daily-stats channel.
3. Note that all posts show a project like count that is greater than the project submissions added count.

### Expected behavior

1. Open up The Odin Project's Discord server.
2. Review recent posts in the top-daily-stats channel.
3. View a count of project likes that is significantly lower than the project submissions added count.

Other

### What OS are you using?

MacOS

### Discord Name

falciron

### Additional Comments

I have a pull request at the ready for this, but would recommend a brief discussion on the issue before it gets reviewed/merged.

If this _bug_ is fixed, we'll see the daily project like count crater from 140+ to 10-20. I personally believe this is the right move, since it provides a more accurate measure of project member activity. I am not aware if the current metric is being reported in any promotional/sponsorship material, however. (But then, I suspect there must be a better source of the data than a daily post in a Discord channel.)

Falciron:
## Because
The daily metrics posted to the Odin Project's #top-daily-stats Discord channel are primarily meant to portray daily **member** activity. Since the single automatic project like granted upon project submission is performed by the system and not a human, it should be omitted from the count of projects liked.

## This PR
- Modifies the project like count method within the daily summary notification service to subtract the number of projects submitted that the same day.

## Issue
Related to #5300

## Additional Information
[Discord Discussion](https://discord.com/channels/505093832157691914/505093832157691916/1496686409862484130)

## Pull Request Requirements
    - `Fix` - bug fixes

### DIFF
--- a/app/services/notifications/daily_summary.rb
+++ b/app/services/notifications/daily_summary.rb
@@ -29,7 +29,7 @@ module Notifications
     end
 
     def project_like_count
-      Like.liked_on(Time.zone.yesterday).count
+      Like.liked_on(Time.zone.yesterday).count - project_submission_count
     end
   end
 end


KevinMulhern (changes requested):
Nice spot with this @Falciron! this was definitely something we overlooked.

Although we'd probably be ok with assuming a 1:1 relationship between submissions and auto-likes, it's not a given. People can unlike their submission, or future maintainers might remove the auto-like feature etc. Then the count could become wrong again silently.

Theres also a potential for negative counts. For example, if 20 projects are submitted and 5 of those learners unliked their submission for whatever reason the formula would be: 15 likes - 20 project submissions.

To make this more robust, I think we should filter out the auto likes at a query level to have the organic like count. I think the approach might be to mark the appropriate likes as auto-liked with a flag:

1. Add an `auto_liked` flag column to the likes table
2. Set the `auto_liked` value to true when we [create the auto like](https://github.com/TheOdinProject/theodinproject/blob/main/app/controllers/lessons/project_submissions_controller.rb#L32) - we should be able pass it through to our like method and set it there.
3. Use a scope like `not_auto_liked` to filter the Likes when counting them in the daily summary.
4. Backfilling `auto_liked` on our legacy like records would be nice too, but that could be left for a separate story.

Falciron:
You raise excellent points, and I think you're especially wise to run the count off of filtered data itself rather than make another short-sighted correction. Do you see the system user changing ID, that we'd need the auto-like to be more than just a generated column? (And if it doesn't change ID, do we need the extra column at all?)

It may take me a little longer to design and implement the correct fix, as I'm still early on in the Ruby path. That said, I can see exactly which table column we'd need for the filter, so it gives me some hope I could figure it out. I'll give it a few days and if I can't figure it out, I'll report back and release the issue if needed.

KevinMulhern:
Thanks @Falciron - no rush, and feel free to ping me with any questions!

> Do you see the system user changing ID, that we'd need the auto-like to be more than just a generated column? (And if it doesn't change ID, do we need the extra column at all?)

We don't have a single system user; we just [programmatically have users like their own projects upon submission](https://github.com/TheOdinProject/theodinproject/blob/main/app/controllers/lessons/project_submissions_controller.rb#L32C11-L32C49). Because the `user_id` varies, filtering isn't straightforward sadly.

We _could_ derive them with a join (`like.user_id == project_submission.user_id`), but I think setting a boolean flag at creation would be cleaner.  It makes the "auto-like" an explicit domain concept and keeps the logic much simpler than relying on data relationships.

PR #5303, opened 2026-04-25
Fix: Success Stories - Remove a broken link of Sean's Szurko site

bartek8b:
The linked page isn't working (ERR_NAME_NOT_RESOLVED)

I've checked on my Windows, Linux and Android devices.

As I can see, the story of Olivia has no linked website so I've used the same approach and set social_media_link to an empty string.


wise-king-sullyman (approved):
Thanks!

PR #5356, opened 2026-05-28
Fix: Fix desktop sidebar padding

### LINKED ISSUE #5364: Bug: Sidebar progress bar missing padding on desktop
### Checks

### Bug description

The progress bar in the lesson sidebar stretches to the right edge on desktop with no padding. 
On mobile the drawer sidebar correctly has `px-4`, but the desktop version is missing it.

**File affected:** `app/components/lessons/sidebar_component.html.erb`

**Current code:**
```html
<div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20">
```

Image

**Expected code:**
```html
<div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 px-4">
```

Image

### How to reproduce

1. Go to any lesson page on a wide desktop screen
2. Look at the progress bar in the left sidebar below the course title
3. The progress bar stretches to the right edge of the sidebar with no padding

### Expected behavior

The progress bar should have padding on both sides, consistent with the mobile 
drawer which correctly has `px-4`.

Chrome

### What OS are you using?

MacOS

### Discord Name

_No response_

### Additional Comments

**Tested fix:** Running the following in the browser DevTools console fixes the issue:

```js
document.querySelector('.sticky.top-0').style.paddingRight = '16px'
document.querySelector('.sticky.top-0').style.paddingLeft = '16px'
```

mao-sz:
## Because

In most cases, seems fine since we have block padding in the sidebar itself and visually some left "padding" (which is from the containing div) but the lack of right padding is very noticeable with the progress bar or highlighted lesson.
In the mobile sidebar, since it's a standalone element, it has its own padding on all 4 sides.

## This PR

- Fixes padding in sidebar component
- Reduces container padding to match
- Simplifies sidebar toggle CSS to prevent horizontal overflow issues

## Issue

Closes #5364

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/app/components/lessons/sidebar_component.html.erb
+++ b/app/components/lessons/sidebar_component.html.erb
@@ -21,10 +21,10 @@
 <%# Static desktop sidebar — xl+ %>
 <aside class="hidden xl:block xl:h-full" data-test-id="lesson-sidebar">
   <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 pr-2">
-    <div class="mb-4 flex flex-col gap-2">
+    <div class="mb-4 px-2 flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
-      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline px-2' %>
+      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline' %>
 
       <% if progress_percentage %>
         <%= render Lessons::Sidebar::ProgressBarComponent.new(percentage: progress_percentage) %>
--- a/app/components/lessons/sidebar_component.html.erb
+++ b/app/components/lessons/sidebar_component.html.erb
@@ -1,9 +1,9 @@
 <%= render Overlays::DrawerComponent.new(hook_class: 'lesson-sidebar-drawer', breakpoint: :xl, close_label: 'Close course contents', aria_label: 'Course contents') do %>
   <div class="flex grow flex-col gap-y-4 overflow-y-auto scrollbar-thin bg-white dark:bg-gray-900 px-4 pb-6 pt-5">
-    <div class="flex flex-col gap-2">
+    <div class="px-2 flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
-      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline px-2' %>
+      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline' %>
 
       <% if progress_percentage %>
         <%= render Lessons::Sidebar::ProgressBarComponent.new(percentage: progress_percentage) %>
--- a/app/components/lessons/sidebar_component.html.erb
+++ b/app/components/lessons/sidebar_component.html.erb
@@ -20,7 +20,7 @@
 
 <%# Static desktop sidebar — xl+ %>
 <aside class="hidden xl:block xl:h-full" data-test-id="lesson-sidebar">
-  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 pr-2">
+  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 px-2">
     <div class="mb-4 px-2 flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
--- a/app/components/lessons/sidebar_component.html.erb
+++ b/app/components/lessons/sidebar_component.html.erb
@@ -20,7 +20,7 @@
 
 <%# Static desktop sidebar — xl+ %>
 <aside class="hidden xl:block xl:h-full" data-test-id="lesson-sidebar">
-  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20">
+  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 pr-4">
     <div class="mb-4 flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
--- a/app/components/lessons/sidebar_component.html.erb
+++ b/app/components/lessons/sidebar_component.html.erb
@@ -3,7 +3,7 @@
     <div class="flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
-      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline' %>
+      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline px-2' %>
 
       <% if progress_percentage %>
         <%= render Lessons::Sidebar::ProgressBarComponent.new(percentage: progress_percentage) %>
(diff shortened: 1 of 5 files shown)


KevinMulhern:
Nice catch @mao-sz!

There seems to be more padding on the right than the left, which is weird because they both show up as 16px in the inspector.

You can see it when comparing the top and right padding. Both p-4's (16px), but the right padding is thicker than the top:
Screenshot 2026-05-29 at 02 54 39


I think it's the overflow for the scroll area. It's reserving space for the scrollbar and adding to the right padding width - roughly 10px.

Changing to pr-1.5 (6px) seems to get it much closer:
Screenshot 2026-05-29 at 03 22 22

mao-sz:
Ugh, great. Forgot to check it on Chrome which is why - the scrollbar gutter is huge in Chrome vs Firefox. Here's how `pr-1.5` looks in FF, in my opinion way too narrow:
image

So it looks like FF uses an overlay scrollbar (checked with standard width) here but Chrome uses a classic scrollbar (with no option for overlay scrollbars - experimental feature removed).

What do you think about keeping `pr-4` but removing `scrollbar-gutter: stable` from the `scrollbar-thin` utility? I think I prefer having the element padding consistent and the contents narrow slightly when the scrollbar appears rather than have a stable gutter than introduces inconsistencies between browsers. The content narrowing isn't *that* intrusive IMHO (but that's definitely highly subjective).

Honestly I just wish there was a standard CSS property `scrollbar-type` with `classic` or `overlay` options. Rather than having to deal with different browsers making them different sizes with different document flow rules and all that ridiculousness.

KevinMulhern:
The layout shift without `scrollbar-gutter: stable` is a little janky. But I've been down a bit of a rabbit hole with this and there doesn't appear to be many good options.

What do you think of keeping `pr-4` and changing to `scrollbar-gutter: stable both-edges;` so theres an equal amount of spacing on both sides to make it look a bit more centered/uniform?

Screenshot 2026-05-30 at 12 44 25

It's possibly too much space around the sidebar content. But I've been staring at this too long to be sure. How does it feel to you?

> Honestly I just wish there was a standard CSS property scrollbar-type with classic or overlay options. Rather than having to deal with different browsers making them different sizes with different document flow rules and all that ridiculousness.

Big agree! or even a way to detect which is enabled so padding could be optimised for it.

mao-sz:
@KevinMulhern I like the `stable both-edges` approach, didn't even cross my mind it'd be suitable. I do think the total space ends up a little too large on Chrome so I tinkered with reducing the padding for that and the containing div in xl from 4 to 2, but then it introduced a horizontal scrollbar....which I located to originate from the sticky sidebar toggle and negative margin shenanigans.

Let me know what you think of the new changes. For me, the spacing is acceptable and more importantly even on both Chrome and Firefox, and there's no horizontal scrollbar stuff as a result. I've brought the course name in left-alignment with the nav too. The sticky toggle also looks and behaves fine at all breakpoints from my testing in both browsers, hopefully with the benefit of simplified classes. If we really want the toggle to be a little further "in" from the edge when the sidebar is collapsed, I can restore the `group` class but with a positive margin or padding this time.

app/components/lessons/sidebar_component.html.erb
```diff
@@ -20,11 +20,11 @@
 
 <%# Static desktop sidebar — xl+ %>
 <aside class="hidden xl:block xl:h-full" data-test-id="lesson-sidebar">
-  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20">
+  <div class="sticky top-0 h-screen overflow-y-auto scrollbar-thin pt-4 pb-20 pr-2">
```
KevinMulhern: Nit: the alignment with this and the back link + progressbar looks a little off. Could we moving the padding up to the wrapping `<div class="mb-4 flex flex-col gap-2">` so the whole header block aligns with the section/lesson rows?

app/views/lessons/_sub_navbar.html.erb
```diff
@@ -1,11 +1,11 @@
 <div
-  class="sticky -top-px z-20 bg-white dark:bg-gray-900 -mx-4 sm:-mx-6 lg:-mx-8 xl:-mr-4 xl:-ml-10 group-[.sidebar-collapsed]/sidebar-layout:xl:-ml-4"
+  class="sticky -top-px z-20 bg-white dark:bg-gray-900"
   data-controller="sticky-state"
   data-sticky-state-stuck-class="border-b border-gray-200 dark:border-gray-800">
-  <div class="h-12 flex items-center px-4 sm:px-6 lg:px-8">
```
KevinMulhern, line 5: Nit: IIRC this padding was matching the main navbar's so the toggle button lined up vertically with the logo. Without it they're visibly offset.

Image

app/views/lessons/_sub_navbar.html.erb
```diff
@@ -1,11 +1,11 @@
 <div
-  class="sticky -top-px z-20 bg-white dark:bg-gray-900 -mx-4 sm:-mx-6 lg:-mx-8 xl:-mr-4 xl:-ml-10 group-[.sidebar-collapsed]/sidebar-layout:xl:-ml-4"
```
KevinMulhern, line 2: I think all of this madness was to "override" the main column's padding and make the sub-navbar's border span the full width. The border will now stop at the parent column's padding edges :(

Image

Is there a simpler way we could handle the padding and still have the border span the full width?

KevinMulhern (commented):
The spacing looks good to me @mao-sz 💪 

Just a few little nit's, the padding on this page is causing us a lot of pain. I wonder if it might be worth localising it to each column instead of the wrapper handling it 🤔

app/components/lessons/sidebar_component.html.erb
mao-sz: Lol I completely forgot to test this with me signed in, so I completely missed the progress bar alignment. Yep, this would unify everything nicely.

app/views/lessons/_sub_navbar.html.erb
mao-sz: That's fair and should be easily manageable.

app/views/lessons/_sub_navbar.html.erb
mao-sz: ~~Probably could just shove it in a div and go from there.~~ Or not

mao-sz:
I think one of the issues is that the parent div of the sidebar and main contents has the inline padding, and so we're having to "match" that padding in the inner elements. That parent div also has a gap between the sidebar and main contents, hence the negative margin shenaningans required.

I think we can ditch the parent padding and gap, and let the sidebar and main contents handle their own padding. That should simplify a lot of the stuff we're messing with right now.

mao-sz:
Right, let's see if this finally covers all bases.

Tested in Chrome and FF on Linux, ensured:
1. Sidebar inline padding is uniform for all elements of it
2. No layout shift when scrollbar is rendered (`stable both-edges`)
3. Sidebar toggle button aligned with logo image in header when sidebar collapsed (all breakpoints)
4. sub_navbar, when sticky, extends full width of viewport (or up the the expanded sidebar in xl). No gaps on the side
5. Expanding/collapsing the sidebar in xl does not shift the lesson contents positioning (ensured works from the breakpoint up to 21:9 ultrawide
6. There is always inline padding around the main lesson contents (inc. lesson ToC in xl)

app/components/lessons/sidebar_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="flex flex-col gap-2">
       <%= render Ui::BackLinkComponent.new(path: path_course_path(course.path, course), name: 'Back to course') %>
 
-      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline' %>
+      <%= link_to course.title, path_course_path(course.path, course), class: 'text-lg font-semibold text-gray-800 dark:text-gray-200 no-underline hover:underline px-2' %>
```
KevinMulhern: Nit: Can we move the px-2 to the wrapper here too please?

app/views/lessons/show.html.erb
KevinMulhern: Nit: What do you think of `xl:w-80` here? With the left page padding gone, the overall width of the sidebar has decreased a little, making it look too narrow perhaps?

If changing is more difficult than it looks, or has too many knock on effects. I think we should keep it at 72 though.

KevinMulhern (commented):
This is looking solid @mao-sz 💪 

Only remaining thing is to update `_sidebar_skeleton.html.erb`  to match the new padding - it still has pr-4 only, so the placeholder bars sit flush-left and shift rightward when the real sidebar loads in.

app/components/lessons/sidebar_component.html.erb
mao-sz: Oops

app/views/lessons/show.html.erb
mao-sz: Simple change since it's just there and `<main>` that need to match to prevent layout shift on toggle. It's only marginal but I *think* I slightly prefer the 80, so let's go with that.

KevinMulhern (approved):
This is looking perfect! Thanks for all the work you put into this @mao-sz ❤️

PR #5366, opened 2026-05-30
Chore: Remove unused PathPrerequisite model

KevinMulhern:
Because:
The PathPrerequisite feature has been dormant since the All Paths page redesign in 2022 removed the UI that displayed prerequisites. Nothing in the app reads `path.prerequisites` or `path.path_prerequisites` anymore; seed files were still writing join rows that no view, controller or service consults.

This PR:
* Remove the `has_many :path_prerequisites` and `has_many :prerequisites` associations from Path
* Delete the PathPrerequisite model, factory and spec
* Drop the prerequisite write lines from the Full Stack Rails and Full Stack JavaScript seed files
* Add a migration to drop the `path_prerequisites` table


db/schema.rb
```diff
@@ -10,25 +10,25 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_02_15_230748) do
+ActiveRecord::Schema[8.1].define(version: 2026_05_30_124754) do
```
KevinMulhern, line 13: Most of these schema changes are included because this is the first time db:migrate was run since the Rails 8.1 bump.

PR #5382, opened 2026-06-06
Feature: Drawer component closes on outside clicks

### LINKED ISSUE #5370: Feature Request: Close sidebar when outside of the sidebar container
### Checks

- [x] The title of this issue follows the `Feature Request: brief description of feature request` format, e.g. `Feature Request: Add a dark mode to the website`

On screen sizes `1024 pixels` and smaller when I have the sidebar open, I'd like to be able to click outside of the sidebar container to close the sidebar instead of clicking the `X` button.

### Acceptance criteria

- [ ] Clicking outside of the sidebar `container` on screen sizes `1024px` and smaller will close the sidebar.

### Additional Comments

_No response_

BryanF1nes:
## Because

Course contents sidebar now closes on outside clicks on smaller screens.

## This PR

- Set pointer events to none on line 23 `class="fixed inset-0 flex pointer-events-none"`
- Set pointer events to auto on line 28 `class="relative mr-16 flex w-full max-w-xs flex-1 hidden pointer-events-auto"`

## Issue

Closes #5370 

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/app/components/overlays/drawer_component.html.erb
+++ b/app/components/overlays/drawer_component.html.erb
@@ -20,7 +20,9 @@
     data-transition-leave-end="opacity-0">
   </div>
 
-  <div class="fixed inset-0 flex">
+  <div 
+    class="fixed inset-0 flex"
+    data-action="click->visibility#backdropClick">
     <div
       class="relative mr-16 flex w-full max-w-xs flex-1 hidden"
       data-visibility-target="content"
--- a/app/components/overlays/drawer_component.html.erb
+++ b/app/components/overlays/drawer_component.html.erb
@@ -20,9 +20,10 @@
     data-transition-leave-end="opacity-0">
   </div>
 
-  <div 
+  <div
     class="fixed inset-0 flex"
     data-action="click->visibility#backdropClick">
+
     <div
       class="relative mr-16 flex w-full max-w-xs flex-1 hidden"
       data-visibility-target="content"
--- a/app/components/overlays/drawer_component.html.erb
+++ b/app/components/overlays/drawer_component.html.erb
@@ -20,12 +20,9 @@
     data-transition-leave-end="opacity-0">
   </div>
 
-  <div
-    class="fixed inset-0 flex"
-    data-action="click->visibility#backdropClick">
-
+  <div class="fixed inset-0 flex pointer-events-none">
     <div
-      class="relative mr-16 flex w-full max-w-xs flex-1 hidden"
+      class="relative mr-16 flex w-full max-w-xs flex-1 hidden pointer-events-auto"
       data-visibility-target="content"
       data-transition-enter="transition ease-in-out duration-200 transform"
       data-transition-enter-start="-translate-x-full"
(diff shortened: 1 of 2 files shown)


app/components/overlays/drawer_component.html.erb
KevinMulhern, line 13: I was refreshing myself on the structure of this component and noticed we have this. We must have had click outside behaviour on this before, and it either stopped working or never worked to begin with lol.

But it got me thinking about why it isn't working - I think it's a stacking issue. This div and the div on line 23 both have `fixed inset-0`, so they occupy the same space in the browser. Since neither has a higher z-index, the browser falls back to DOM order to decide what's on top - last one wins, so the line 23 div sits above this one and absorbs the clicks before they reach it 

Since `visibility_controller` is shared across a few components, I’d love to avoid adding a new method there if we can, just to keep its API small and generalised. It seems this might actually be a drawer-specific layout issue rather than missing behaviour (the close action is being blocked). I think we can use a more localised CSS fix with pointer events instead. I'll map it out in the next couple of comments.

Let me know what you think!

app/components/overlays/drawer_component.html.erb
```diff
@@ -20,7 +20,10 @@
     data-transition-leave-end="opacity-0">
   </div>
 
-  <div class="fixed inset-0 flex">
+  <div
+    class="fixed inset-0 flex"
```
KevinMulhern: We can add `pointer-events-none` here so the click will go through the to the div below in the browser (the one on line 10 because of browser stacking).
 
```suggestion
    class="fixed inset-0 flex pointer-events-none"
```

app/components/overlays/drawer_component.html.erb
```diff
@@ -20,7 +20,10 @@
...
+    data-action="click->visibility#backdropClick">
+
     <div
       class="relative mr-16 flex w-full max-w-xs flex-1 hidden"
```
KevinMulhern: Now we can re-enable clicks again so the nav items and close button continue to work:
 
```suggestion
      class="relative mr-16 flex w-full max-w-xs flex-1 hidden pointer-events-auto"
```

app/components/overlays/drawer_component.html.erb
BryanF1nes: This makes perfect sense, when I was tinkering around with it and seeing how it was intended to work that line there `data-action="click->visibility#off"` was making me think the same thing that at one point the intended functionality was what I was proposing.

I did come to the same conclusion before sending in the PR that it was a stacking issue, I had created a helper function inside the visibility controller to see what exactly was being clicked and it was never the DIV element I was expecting it to be, which lead me to a stacking issue.

This makes sense why you would want to keep the API small! I am 100% on board with that decision of using CSS to fix the pointer events instead of making the API any bigger than it needs to be. I will apply your suggestions, thanks!

KevinMulhern (approved):
QA'd across main nav, lesson sidebar, and admin main nav - all working great ✅ 

Nice work @BryanF1nes!

PR #5395, opened 2026-06-12
Fix: Change Lesson Buttons Grid Breakpoint

### LINKED ISSUE #5384: Bug: Lesson navigation buttons overlap Mark Complete button at certain viewport widths
### Checks

### Bug description

On lesson pages, the Previous Lesson and Next Lesson navigation buttons overlap the Mark Complete button at certain viewport widths before the responsive breakpoint is reached.

The issue occurs when the browser width is approximately 1280px–1380px. Instead of maintaining proper spacing, the navigation buttons encroach into the center area occupied by the Mark Complete button, causing the controls to overlap visually.

Screenshot attached showing the overlap.

Image

### How to reproduce

1. Open any lesson page.
2. Scroll to the bottom of the lesson.
3. Resize the browser window to a width between approximately 1280px and 1380px.
4. Observe the lesson navigation controls.

### Expected behavior

The Previous Lesson, Mark Complete, and Next Lesson buttons should remain properly spaced and not overlap at any viewport width.

Chrome

### What OS are you using?

Ubuntu

### Discord Name

_No response_

### Additional Comments

_No response_

ZackHoang:
## Because

-  Buttons at the end of lesson are overlapping on each other at around 1280px-1380px. This was actually due to breakpoints for the button's `<div>` being set too small which causes some overlap. The most straight-forward fix here would be to use tailwind's default breakpoint `2xl` to maintain the button's flex at higher screen widths

## This PR

-  Changes the button's <div> breakpoint from md (48rem = 768px) to 2xl (96rem = 1536px)

## Issue

Closes #5384 

## Pull Request Requirements

    - `Fix` - bug fixes

### DIFF
--- a/app/views/lessons/_lesson_buttons.html.erb
+++ b/app/views/lessons/_lesson_buttons.html.erb
@@ -1,9 +1,9 @@
-<div class="flex flex-col gap-y-6 max-w-sm mx-auto 2xl:grid 2xl:grid-cols-3 2xl:gap-x-7 2xl:gap-y-0 2xl:max-w-full items-center">
+<div class="flex flex-col gap-y-6 max-w-sm mx-auto md:grid md:grid-cols-[1fr_auto_1fr] md:gap-x-7 md:gap-y-0 md:max-w-full md:items-center">
   <% if course.previous_lesson(lesson).present? %>
     <div class="contents md:block md:justify-self-start">
-      <%= link_to lesson_path(course.previous_lesson(lesson)), class: 'button button--secondary px-4 py-2 min-w-45 font-medium', data: { test_id: 'previous-lesson-btn' } do %>
+      <%= link_to lesson_path(course.previous_lesson(lesson)), class: 'button button--secondary px-4 py-2 font-medium', data: { test_id: 'previous-lesson-btn' } do %>
         <%= inline_svg_tag 'icons/arrow-left-circle.svg', class: 'h-8 w-8 pr-2', aria: true, title: 'Previous lesson', desc: 'Go to previous lesson' %>
-        Previous Lesson
+        Previous
       <% end %>
     </div>
   <% else %>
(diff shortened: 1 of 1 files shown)


app/views/lessons/_lesson_buttons.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="flex flex-col gap-y-6 max-w-sm mx-auto md:grid md:grid-cols-3 md:gap-x-7 md:gap-y-0 md:max-w-full md:items-center">
+<div class="flex flex-col gap-y-6 max-w-sm mx-auto 2xl:grid 2xl:grid-cols-3 2xl:gap-x-7 2xl:gap-y-0 2xl:max-w-full items-center">
```
KevinMulhern: Thanks for digging into this @ZackHoang! The change does stop the overlap, but I think it over-corrects a bit. Moving the grid to 2xl means the three buttons now stack vertically on
everything below 1536px - so most laptop screens (1280–1440px) lose the side-by-side row, not just the narrow band where the overlap actually happened.

Image

The root cause is that the lesson page's two sidebars both appear at xl (1280px), which squeezes the content column down to ~550px. The buttons can't fit three-across there because they're pinned wider than the space: the side buttons have a min-w-45 (180px) floor and the middle "Mark Complete" is a fixed w-60 (240px).

So a more targeted fix would keep the row and just let the buttons fit the squeeze, e.g.:
- swap md:grid-cols-3 → md:grid-cols-[1fr_auto_1fr] (centers the middle button and gives it its own track):
```suggestion
<div class="flex flex-col gap-y-6 max-w-sm mx-auto md:grid md:grid-cols-[1fr_auto_1fr] md:gap-x-7 md:gap-y-0 md:max-w-full md:items-center">
```

  - drop min-w-45 from the prev/next buttons,
  - and shorten the labels to  from "Previous lesson"/"Next lesson" to just "Previous"/"Next" (the arrows already say which direction and we don't need to repeat "lesson" so much in this context).

That clears the overlap while keeping the side-by-side layout at all widths.

KevinMulhern (changes requested):

ZackHoang:
@KevinMulhern Made the changes you asked in a new commit! I can see that my old changes are kinda quick and dirty. I also thought we should have kept the button rows on laptop screen's resolution but I didn't think of it that way. Thanks for the suggestion! :smile:

KevinMulhern (approved):
LGTM, thanks for doing this @ZackHoang!

PR #5407, opened 2026-06-23
Chore: Remove unused postcss dependency and dead config

KevinMulhern:
The postcss.config.js file referenced @tailwindcss/postcss, which isn't even installed, and the Tailwind v4 CLI never actually loaded it — the CSS build produces identical output without it. This removes the dead config file, drops the now-unused --postcss flag from the build:css script, and removes the direct postcss dependency (it stays available transitively through stylelint and tailwindcss).


mao-sz (approved):
:rocket:

PR #5442, opened 2026-07-30
fix: improve lesson content navigation semantics

### LINKED ISSUE #4423: Bug: lesson side navigation a11y issues

WaghuleAsmita:
## Because

The lesson contents sidebar should be exposed as a navigation landmark and properly labeled for assistive technologies, as described in issue #4423.

## This PR

- Replaces the lesson contents wrapper from `<aside>` to `<nav>`
- Adds `aria-labelledby` to the navigation element
- Adds an `id` to the "Lesson contents" heading
- Adds `role="list"` and `aria-labelledby` to the table of contents `<ul>`

## Issue

Closes #4423

### DIFF
--- a/app/views/lessons/show.html.erb
+++ b/app/views/lessons/show.html.erb
@@ -78,13 +78,19 @@
           </div>
         </main>
 
-        <aside class="hidden xl:block xl:w-72 xl:shrink-0">
+        <nav class="hidden xl:block xl:w-72 xl:shrink-0" aria-labelledby="lesson-contents-heading">
           <div class="sticky top-12 pt-10 sm:pt-14 pb-20">
-            <h4 class="text-md pb-4 text-gray-700 dark:text-gray-300">Lesson contents</h4>
-            <ul class="flex flex-col text-gray-500 dark:text-gray-400" data-lesson-toc-target="toc"></ul>
+
+            <h4 id="lesson-contents-heading" class="text-md pb-4 text-gray-700 dark:text-gray-300">
+              Lesson contents
+            </h4>
+
+            <ul class="flex flex-col text-gray-500 dark:text-gray-400" data-lesson-toc-target="toc" role="list" aria-labelledby="lesson-contents-heading">
+            </ul>
           </div>
-        </aside>
+        </nav>
       </div>
     </div>
   </div>
 </div>
+


WaghuleAsmita:
Hi maintainers, I’ve implemented the accessibility improvements described in issue #4423.

The lesson contents sidebar now uses proper navigation semantics with `nav`, `aria-labelledby`, and the appropriate heading/list labeling.

Could you please review the PR when you get a chance? Thank you!

app/views/lessons/show.html.erb
```diff
@@ -78,13 +78,19 @@
...
+        <nav class="hidden xl:block xl:w-72 xl:shrink-0" aria-labelledby="lesson-contents-heading">
           <div class="sticky top-12 pt-10 sm:pt-14 pb-20">
-            <h4 class="text-md pb-4 text-gray-700 dark:text-gray-300">Lesson contents</h4>
-            <ul class="flex flex-col text-gray-500 dark:text-gray-400" data-lesson-toc-target="toc"></ul>
+
+            <h4 id="lesson-contents-heading" class="text-md pb-4 text-gray-700 dark:text-gray-300">
+              Lesson contents
+            </h4>
+
+            <ul class="flex flex-col text-gray-500 dark:text-gray-400" data-lesson-toc-target="toc" role="list" aria-labelledby="lesson-contents-heading">
```
KevinMulhern, line 88: Can `aria-labelledby="lesson-contents-heading"` be dropped from here since it's already on the nav element?

app/views/lessons/show.html.erb
```diff
@@ -78,13 +78,19 @@
...
+            </ul>
           </div>
-        </aside>
+        </nav>
       </div>
     </div>
   </div>
 </div>
+
```
KevinMulhern, line 96: Nit: Can this trailing blank line be removed please?

app/views/lessons/show.html.erb
```diff
@@ -78,13 +78,19 @@
           </div>
         </main>
 
-        <aside class="hidden xl:block xl:w-72 xl:shrink-0">
+        <nav class="hidden xl:block xl:w-72 xl:shrink-0" aria-labelledby="lesson-contents-heading">
```
KevinMulhern, line 81: Nit: I think we use aria-label instead of `aria-labelledby` everywhere else on the app. That would keep this consistent with the rest of the codebas, and allow us to drop the id on the h4.

KevinMulhern (changes requested):

PR #5462, opened 2026-08-21
Chore: Dedupe prismjs (resolve to 1.30.0)

mao-sz:
## Because

Fixes https://github.com/TheOdinProject/theodinproject/security/dependabot/167
We were already using prismjs@1.30.0 but for some reason, `esbuild-plugin-prismjs` was stubbornly clinging to `1.29.0` despite resolving `^1.29.0`. Because of that, #5461 failed CI due to making a `yarn.lock` change that'd trigger Yarn's immutable+hardened mode.

Fix is to run `yarn dedupe prismjs` to remove 1.29.0 and make `esbuild-plugin-prismjs` use 1.30.0.
Local CI passes (only addition in 1.30.0 is a check that's part of the linked CVE).

## This PR

- Dedupes prismjs to only use v1.30.0

## Issue

N/A

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5731,14 +5731,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"prismjs@npm:^1.29.0":
-  version: 1.29.0
-  resolution: "prismjs@npm:1.29.0"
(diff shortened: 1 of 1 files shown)


KevinMulhern (approved):
Nice spot!

